PR #4, opened 2022-03-03
initial pheno job

luciipak:


luciipak:
the directories under /pytorch were basically just copied over. I kept everything, but there a lot of unused files. Should I just go ahead and remove them?

Very open to suggestions and comments on better practices, it is similar to how it was originally running just to get it working. 

still needed: testing, while loop to check for new queue item before shutting down, setup db secrets.

jobs/pheno/pheno-worker/Makefile
```diff
@@ -0,0 +1,21 @@
+ecr_repo = 077346344852.dkr.ecr.us-east-2.amazonaws.com/pheno-worker
+# ecr_repo = lucipak/pheno-job
```
jperson, line 2: can you remove this comment

jobs/pheno/pheno-worker/terraform/main.tf
```diff
@@ -0,0 +1,47 @@
...
+  name                 = "pheno-worker"
+  image_tag_mutability = "MUTABLE"
+
+  image_scanning_configuration {
+    scan_on_push = true
+  }
+}
+
+// Get date from another state file
+data "terraform_remote_state" "state1" {
```
jperson, line 40: you don't need this data resource, I only added it to the sample as an example for getting remote state

jperson (approved):

PR #26, opened 2022-03-14
initial user auth service

jperson:


deployments/apiv2/services/users/src/main.py
```diff
@@ -0,0 +1,131 @@
...
+    await db.close()
+
+
+@app.get("/users/me", response_model=UserProfile)
+async def index(request: Request, token=Depends(ProtectedAny(scope=["users:free"]))):
+    try:
+        async with request.state.pgpool.acquire() as con:
+            user_id = uuid.UUID(hex=token["userid"])
+            rows = await con.fetchrow("select name, email, user_id, account_type, created_at, updated_at, data->'scope' as scope from users where user_id = $1", user_id)
+
```
luciipak, line 51: whats the data->'scope' as scope?

deployments/apiv2/services/users/src/main.py
```diff
@@ -0,0 +1,131 @@
...
+                ph.hash(p)
+                raise LoginError(failed_msg)
+            else:
+                scope = json.loads(row.get("scope", "[]"))
+                iat = timegm(datetime.now(tz=timezone.utc).utctimetuple())
+                exp = timegm((datetime.now(tz=timezone.utc) + timedelta(minutes=ACCESS_TOKEN_EXPIRE_MINUTES)).utctimetuple())
+
+                jwt_meta = JWTMeta(aud=JWT_AUDIENCE, scope=scope, iat=iat, exp=exp)
+                jwt_details = JWTDetails(userid=row["user_id"].hex)
+                jwt_payload = JWTPayload(**jwt_meta.dict(), **jwt_details.dict())
```
luciipak, line 92: also, is the ** in python the same as the spread operator in js -> [...array] ?

deployments/apiv2/services/users/src/main.py
jperson: `data` is a `jsonb` column in the table, so `data->'scope'` is getting the `scope` field from the json object for the `data` column in that row, the `as scope` is setting the output column name to `scope` instead of the something like `?column?` that would be set otherwise

deployments/apiv2/services/users/src/main.py
jperson: yeah pretty much, `**` splats `key=value` for each item in a dictionary, for any array in python you would use a single `*`

you'll see stuff like

```python
def some_func(*args, **kwargs):
    ....
```
in python that spreads the unamed arguments from a list and then all the key word arguments from a dict

luciipak (approved):

PR #30, opened 2022-03-17
test cluster update, added maunits migration

jperson:


core/db/curibio/alembic/versions/1fbfdf12dfbb_add_maunits.py
```diff
@@ -0,0 +1,51 @@
...
+            {"serial_number": "M02022055001", "hw_version": "2.2.0"},
+            {"serial_number": "M02022055002", "hw_version": "2.2.0"},
+            {"serial_number": "M02022055003", "hw_version": "2.2.0"},
+            {"serial_number": "M02022055004", "hw_version": "2.2.0"},
+        ],
+    )
+
+    op.execute("GRANT SELECT ON maunits TO curibio_mantarray_ro")
+
+    mantarray_ro_pass = os.getenv("MANTARRAY_RO_PASS")
```
luciipak, line 40: is this using the sealed secret "curibio_mantarray_ro" in the curibio-mantarray-creds.yaml?

jperson: it will in the CI workflow, I have a .env file I use locally to load all the needed passwords, I think the plan is to run all the migrations from CI. For the test cluster you can read out the secrets with kubectl and put them in a .env if you want to test out migrations on the test db

luciipak (approved):

PR #31, opened 2022-03-18
FW auto updating

tannermpeterson:


deployments/apiv2/manifests/apiv2-dep.yaml
```diff
@@ -0,0 +1,57 @@
...
+    http:
+      paths:
+      # - path: /users
+      #   pathType: Prefix
+      #   backend:
+      #     service:
+      #       name: users
+      #       port:
+      #         number: 80
+      - path: /mantarray/
```
jperson, line 51: I don't' think you need the trailing '/'

deployments/apiv2/services/mantarray/src/main.py
```diff
@@ -0,0 +1,40 @@
...
+async def root(request: Request, cur=Depends(get_cursor("reader"))):
+    cur.execute("SELECT * FROM MAUnits;")
+    results = cur.fetchall()
+    units = [{col: row[col] for col in ("serial_number", "hw_version")} for row in results]
+    return templates.TemplateResponse("table.html", {"request": request, "units": units})
+
+
+@app.get("/firmware_latest")
+async def get_latest_firmware(serial_number: str, cur=Depends(get_cursor("reader"))):
+    cur.execute(f"SELECT hw_version FROM MAUnits WHERE serial_number = '{serial_number}';")
```
jperson, line 27: when you're sending queries with user supplied params you need to use prepared queries, this is a common attack vector for sql injection, if someone were to pass in a query string like `test OR 1 = 1` it would return all the rows from the db instead of the one matching the serial number.

Not really an issue for this table since there isn't any sensitive data but it's a good idea to always use prepared statements when querying a database.

All you have to do is change it to this
```
cur.execute("SELECT hw_version FROM MAUnits WHERE serial_number = '$1'", serial_number)
```
and the database engine will substitute the parameters safely, you also don't need to add ';' to the end of statements when using a database client

jperson: obligatory Bobby tables

tannermpeterson: Ah yes, should've thought about that.

deployments/apiv2/manifests/apiv2-dep.yaml
tannermpeterson: I was running across an issue where if I didn't use the trailing `/` then it would rewrite `/mantarray/firmware_latest` to `//firmware_latest`. Not sure if there is a better way to solve that issue, but this seemed to work

jperson (approved):

PR #46, opened 2022-03-23
removed md5

luciipak:


core/lib/utils/utils/s3.py
```diff
@@ -44,24 +44,15 @@ def generate_presigned_urls_for_dir(bucket: str, key_prefix: str, objs_only: boo
         raise ClientError(f"Failed to generate presigned urls for {bucket}/{key_prefix}: {e}")
 
 
-def generate_presigned_post(bucket: str, key: str, file_path) -> Dict[Any, Any]:
+def generate_presigned_post(bucket: str, key: str, file_path: str, md5s: str) -> Dict[Any, Any]:
```
jperson, line 47: sorry, not trying to be nit-picky, but would we need the `file_path` param, I think that the `key` param would include the file name and path in s3 that we wanted to upload the file to

luciipak: no, that's my bad. I used it in the log statement so I didn't notice it wasn't used elsewhere. changing now!

jperson (approved):

PR #74, opened 2022-04-15
added metadata handling for pulse3d worker

luciipak:


core/db/curibio/alembic/versions/49fceab10078_create_pulse3d_tables.py
```diff
@@ -0,0 +1,131 @@
...
+            ["mantarray_recording_session_id"],
+            ["mantarray_recording_sessions.mantarray_recording_session_id"],
+            ondelete="SET NULL",
+        ),
+    )
+
+    op.create_table(
+        "mantarray_raw_files",
+        sa.Column("well_index", sa.SmallInteger(), nullable=False),
+        sa.Column("upload_id", sa.Integer(), nullable=True),
```
jperson, line 85: we might want a foreign key constraint to the uploads table on this column

jobs/pulse3d/pulse3d-worker/src/lib/queries.py
```diff
@@ -0,0 +1,25 @@
...
+
+INSERT_INTO_MANTARRAY_RECORDING_SESSIONS = """
+    INSERT INTO mantarray_recording_sessions(mantarray_recording_session_id, customer_account_id, user_account_id, instrument_serial_number, session_log_id, acquisition_started_at, length_microseconds,
+    recording_started_at)
+    VALUES ($1, $2, $3, $4, $5, $6, $7, $8);
+    """
+
+INSERT_INTO_MANTARRAY_RAW_FILES = f"""
+    INSERT INTO mantarray_raw_files(well_index, upload_id, length_microseconds, recording_started_at, mantarray_recording_session_id)
+    VALUES ($1, {SELECT_LAST_UPLOAD_ID}, $2, $3, $4);
```
jperson, line 16: we might want to instead use a transaction and only select the last upload id once for all the inserts and then pass it as a normal parameter to the insert queries, depending on how this is used it might cause an issue doing multiple selects to get the last upload

core/db/curibio/alembic/versions/49fceab10078_create_pulse3d_tables.py
```diff
@@ -0,0 +1,131 @@
...
+from sqlalchemy.sql import func
+
+revision = "49fceab10078"
+down_revision = "1fbfdf12dfbb"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
```
jperson, line 20: there's already an uploads table that I added, we might want to use that instead or delete it if we go with this one
https://github.com/CuriBio/k8s_curi/blob/main/core/db/curibio/alembic/versions/258ca806ee8b_initial_jobs_queue_table.py#L27

jobs/pulse3d/pulse3d-worker/src/main.py
```diff
@@ -37,38 +39,50 @@ async def process(con, item):
...
-            return ('error', {})
+            return ("error", {})
 
         try:
             logger.info(f"Starting pulse3d analysis")
             outfile = f"{'.'.join(filename.split('.')[:-1])}.xlsx"
-            prs = PlateRecording.from_directory(f"/{tmpdir}")
-            for r in prs:
-                write_xlsx(r, name=outfile)
+            prs = PlateRecording.from_directory(tmpdir)
```
jperson, line 55: I think for this we might want to do the for loop even though the mantarray will only have 1 recording per zip, when we build out the mybinder replacement users might upload a zip file with multiple recordings in it

jperson (approved):

PR #84, opened 2022-05-10
Firmware bucket prefix

tannermpeterson:


deployments/apiv2/manifests/overlays/prod/kustomization.yaml
```diff
@@ -26,3 +26,8 @@ patchesJson6902:
     - op: replace
       path: /spec/template/spec/containers/1/image
       value: "245339368379.dkr.ecr.us-east-2.amazonaws.com/mantarray"
+    - op: replace
+      path: /spec/template/spec/containers/0/env/0/value
```
jperson, line 30: I think the path should be `/spec/tempate/spec/containers/1/env/0/value` since the `mantarray` container is second in the containers list

jperson:
LGTM

jperson (approved):

PR #102, opened 2022-05-27
Update wworker

luciipak:


frontend/pulse3d/src/utils/worker.js
```diff
@@ -25,13 +27,14 @@ self.onmessage = async ({ data }) => {
 const handleGenericRequest = async ({ method, endpoint, body }) => {
   const url = `${baseUrl}${endpoint}`;
 
-  // add token to request headers
-  const headers = new Headers();
-  headers.append('Authorization', `Bearer ${authToken}`);
-  request.headers = headers;
+  const reqInstance = axios.create({
```
tannermpeterson, line 30: I was actually looking into doing something like this for the desktop app routes so we don't have to write the boiler plate so often

tannermpeterson (approved):

PR #103, opened 2022-06-03
initial uploads table and web worker persisting auth token

luciipak:


core/lib/utils/utils/s3.py
```diff
@@ -14,7 +14,7 @@ class S3Error(Exception):
 
 def generate_presigned_url(bucket: str, key: str, exp: int = 3600) -> Any:
     s3 = boto3.resource("s3")
-    s3_client = boto3.client("s3")
+    s3_client = boto3.client("s3", config=Config(signature_version="s3v4", region_name='us-east-2'))
```
luciipak, line 17: should we make two of these functions unique to region since the desktop app needs us-east-1?

PR #104, opened 2022-06-03
web working persisting token and initial uploads table

luciipak:


tannermpeterson (approved):

deployments/apiv2/services/users/src/main.py
```diff
@@ -23,19 +23,7 @@
 
 asyncpg_pool = AsyncpgPoolDep(dsn=DATABASE_URL)
 
-
 app = FastAPI(openapi_url=None)
-app.add_middleware(
```
luciipak, line 28: this is still in this file, there were just two

PR #109, opened 2022-06-09
Initial Refresh Functionality

tannermpeterson:


frontend/pulse3d/src/utils/worker.js
```diff
@@ -22,7 +35,7 @@ Expected message format:
 
 // message handler
 onmessage = async ({ data }) => {
-  console.log("WW onmessage:", accessToken, data);
```
luciipak, line 25: do we want to keep these logs?

frontend/pulse3d/src/utils/worker.js
```diff
@@ -22,7 +35,7 @@ Expected message format:
...
+  console.log("WW onmessage:", tokens, data);
```
luciipak, line 38: do we want to keep these in?

luciipak (approved):

PR #118, opened 2022-07-01
Reanalysis feature

luciipak:


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -111,6 +117,33 @@ async def create_recording_upload(
...
+
+        meta = {
+            "prefix": f"uploads/{customer_id}/{user_id}",
+            "filename": details.filename,
+            "re_analysis": True,
+            "orig_upload_id": details.upload_id,
+        }
+
+        async with request.state.pgpool.acquire() as con:
+            upload_id = await create_upload(con=con, user_id=user_id, meta=meta)
```
jperson, line 136: Is this method for re-running analysis? If so I don't think we want a new upload id.

luciipak: @jperson I was thinking about that, but then do we want the new job to update the db fields for the original analysis? and then in the web app, does matter if users can only  see and download the most recent run since there will only be the one upload id? I can go in and add db handling for updating instead of inserting on reanalysis

jperson: I think the analysis only changes the jobs_result table, the metadata in the uploads is the md5, prefix, and file name. We can remove the object_key from the uploads table and add it to the job_result table, then in the ui we can group all the jobs output with each upload id.

luciipak: Okay cool, yeah  that makes a lot of sense and would be a lot easier.. I'll make the migration. 
Also, it would be nice to eventually separate or figure out a better way of returning presigned urls when getting jobs.. it takes a really long time if a user has run a bunch of jobs. More just a thought out loud so I don't forget that. I think it would mess up in the desktop app if I did that now.. so was thinking of adding that after the release for v1 since it is only slowing down the web app.

luciipak: @jperson Could I also  add the meta keys as columns in the uploads table? Because there's already an md5 column that isn't getting used and I actually use the prefix and filename get used a lot and its kind of an extra step to parse it.

jperson: yeah I think that makes sense, when you do the migration it's probably fine to delete all the rows from the uploads instead of migrating the existing data

jperson (approved):

PR #123, opened 2022-07-09
minor fix to support safari

luciipak:


frontend/pulse3d/src/components/layouts/DashboardLayout.js
```diff
@@ -20,16 +20,24 @@ export default function DashboardLayout({ children }) {
 
   useEffect(() => {
```
tannermpeterson, line 21: Is this `useEffect necessary anymore now that the one beneath it got added?

tannermpeterson (approved):

frontend/pulse3d/src/components/layouts/DashboardLayout.js
luciipak: Yeah I'm trying to think of a better way of doing it, but I need it to get called everything the component mounts, when the account type changes, and when the router changes.

luciipak: because if a user refreshes the page, then it needs to happen on mount too

tannermpeterson: Gotcha

PR #132, opened 2022-07-18
Add s3 artifacts for argo-workflows

jperson:


cluster/manifests/argo-workflows/10-argo_workflows_namespace_install_v3.2.9.yaml
```diff
@@ -667,11 +667,11 @@ spec:
...
----
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: argo
+# ---
+# apiVersion: v1
+# kind: ServiceAccount
+# metadata:
+#   name: argo
```
jperson, line 674: leaving this as comments for now since I might need to put it back

tannermpeterson (approved):

cluster/terraform/k8s/json/argo_namespace_iam_policy.json
```diff
@@ -16,6 +16,14 @@
...
                 "*"
             ]
+        },
+        {
+          "Effect": "Allow",
+          "Action": ["s3:PutObject", "s3:GetObject"],
+          "Resource": [
+            "arn:aws:s3:::curi-prod-workflows",
+            "arn:aws:s3:::curi-prod-workflows/*"
+          ]
```
jperson, line 26: I already ran the terraform to create the buckets, this is only for prod so there isn't a test overlay

cluster/terraform/k8s/modules/argo_workflows/main.tf
```diff
@@ -0,0 +1,18 @@
+resource "aws_s3_bucket" "workflow_artifacts" {
+  bucket = "curi-${var.cluster_name}-workflows"
+}
```
jperson, line 3: it will create a bucket in the test env, I was being lazy about making it only in prod but it won't hurt anything to have an empty bucket in test

PR #155, opened 2022-07-24
CI terraform plan apply

luciipak:


cluster/manifests/ci/base/templates/build-container-template.yaml
```diff
@@ -34,6 +34,34 @@ spec:
 
         SERVICES=$(builder --changed --status=pending --sha={{inputs.parameters.sha}})
         echo $SERVICES
+  - name: checkout-tf
```
jperson, line 37: I don't think you need to do a git checkout since you have the volume mount that the first git checkout used to clone the repo

luciipak: @jperson yeah that's true. I just changed it to grab the terraform changes, not checkout or fetch.

cluster/manifests/ci/base/templates/build-container-template.yaml
```diff
@@ -34,6 +34,28 @@ spec:
...
+          mountPath: /tmp/git
+      env:
+        - name: TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: github-access-token
+              key: token
+      command: ["sh"]
+      source: |
+        cd k8s_curi
```
jperson, line 55: just a nitpick, but you could set the above `workingDir` to `/tmp/git/k8s_curi` instead

jperson (approved):

PR #156, opened 2022-07-26
added ./jobs to git diff

NikitaPoly:


jperson:
I tested this out locally by making a new branch with changes in deployments/pulse3d and jobs/pulse3d. Since the directory structure for `jobs` is slightly different than `deployments` I think this will need to take that into account.

```
In [5]: find_changed('fe5e4078648a041e7d07750873cdbcc886b0bad8')
Out[5]:
[{'path': './deployments/pulse3d/services/pulse3d',
  'deployment': 'pulse3d',
  'service': 'pulse3d_api'},
 {'path': './jobs/pulse3d/pulse3d-worker',
  'deployment': 'pulse3d',
  'service': 'src'}]
```

For a changed job the service gets set to `src`, I think it should be `pulse3d-worker` instead

bin/curi_builder/builder/__init__.py
```diff
@@ -12,22 +12,23 @@
...
     return [
         {
-            "path": f"./{'/'.join(x.split('/')[:-2])}",
-            "deployment": x.split("/")[1],
-            "service": x.replace("/services/pulse3d", "/services/pulse3d_api").split("/")[
-                3
-            ],  # leaving services to prevent switching for pulse3d-worker
+            "path": f"./{'/'.join(ch.split('/')[:-2])}",
+            "deployment": ch.split("/")[1],
+            "service": "pulse3d_api" if "deployments" in ch else "pulse3d-worker"
```
jperson, line 28: this won't work, there are other services in deployments besides pulse3d_api and in the future there will be other jobs besides pulse3d-worker

jperson (approved):

PR #164, opened 2022-08-08
added alert for logged out

NikitaPoly:


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -47,6 +47,7 @@ class JobRequest(BaseModel):
     allow_origins=[
         "https://dashboard.curibio-test.com",
         "https://dashboard.curibio.com",
+        "http://localhost:3000"
```
luciipak, line 50: you'll want to remove these

luciipak (approved):

PR #168, opened 2022-08-11
added advanced params and handling

NikitaPoly:
-- Added Advanced params to FE
-- Changed BE to allow for new types of params
-- Added styles for analysis page


luciipak:
Can we clean up the styling a bit? Base it off how it was originally.
Make all the inputs the same width, move the checkbox headers over the borders, put all the labels on one side and space everything equally. You could probably duplicate the windowed analysis component and swap out the labels.

Screen Shot 2022-08-11 at 3 36 24 PM
Screen Shot 2022-08-11 at 3 36 34 PM

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -338,56 +404,152 @@ export default function AnalysisParamForm({
...
+          <WAOverlayContainer>
+            {checkedAdvanced || <WAOverlay />}
+            <TwoParamContainer>
+              <Label htmlFor="prominenceFactorMin">Prominence (uN):</Label>
+              <InputErrorContainer>
+                <label htmlFor="prominenceFactorMin">min</label>
+                <FormModify>
+                  <FormInput
+                    name="prominenceFactorMin"
+                    placeholder={checkedAdvanced ? "0" : ""}
```
luciipak, line 478: and I would make the placeholder the default values that get used in Pulse3D

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -338,56 +404,152 @@ export default function AnalysisParamForm({
...
+                    <ErrorText id="prominenceFactorMaxError" role="errorMsg">
+                      {errorMessages.prominenceFactorMax}
+                    </ErrorText>
+                  </FormInput>
+                </FormModify>
+
+              </InputErrorContainer>
+            </TwoParamContainer>
+            <TwoParamContainer>
+              <Label htmlFor="widthFactorMin">Width (ms):</Label>
```
luciipak, line 512: and is this ms? I don't think ms is used anywhere

tannermpeterson: Hamed wants us to use ms here, but I'm pretty sure the scipy peak finding function takes the value as a number of samples, so somewhere in the server we need to convert `# samples -> ms`. https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.find_peaks.html

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
tannermpeterson: I agree, and depending on what our default values are we'll need to handle the max/min combos carefully. For both prominence and width, scipy will let us enter either None, a min only, or a max and min. We can't enter just a max which is unfortunate. So for example, one case that's tricky to decide what to do is if the user enters only a max and the value they choose is lower than the default min. We should probably discuss this more in standup or set some time aside for it. Might also be good to see what Hamed thinks

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -338,56 +404,152 @@ export default function AnalysisParamForm({
...
+                      {errorMessages.widthFactorMin}
+                    </ErrorText>
+
+                  </FormInput>
+                </FormModify>
+                <label htmlFor="widthFactorMax">max</label>
+                <FormModify>
+                  <FormInput
+                    name="widthFactorMax"
+                    placeholder={checkedAdvanced ? "100" : ""}
```
tannermpeterson, line 536: If there's no default max in the BE/pulse3d then we'll want to use None/null here

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -35,8 +35,9 @@ class UploadResponse(BaseModel):
 
 class JobRequest(BaseModel):
     upload_id: uuid.UUID
-    prominence_factors: Optional[Union[Tuple[Union[int, float]], List[int]]]
-    width_factors: Optional[Union[Tuple[Union[int, float]], List[int]]]
+    prominence_factors:Optional[Union [Tuple [Union [int,float]], List[Union[int,float]],int,float]]
```
tannermpeterson, line 38: Since we don't have pre-commit in this repo it might be good to install black to format python files and prettier to format js files. If you're using VSCode you can install both of these as VSCode extensions and configure them to format on save

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
luciipak: I think it’s defaulting to 7 for both in pulse3D and prominence is defaulting to 6 for both, but yeah might be nice to make sure it all works with None/null

tannermpeterson: Now that I'm looking through the pulse3d code again, the two values in the parametrization aren't actually a max and min, they're factors used in determining the min value for peaks and the min value for valleys respectively. I think we'll definitely need to discuss more with Jason and Hamed tomorrow

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -290,6 +299,25 @@ async def create_new_job(
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
 
+def _format_advanced_options(option: List[Union[int, float, None]], optionName):
+    if option is None:
+        return None
+    # if only peaks is passed return tupele(peaks,)
```
tannermpeterson, line 305: typo

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -143,65 +171,125 @@ export default function AnalysisParamForm({
...
     }
     setAnalysisParams(updatedParams);
   };
 
-  const validateTuple = (userInput) => {
-    const tuple = userInput.split(",")
-    //check there are only two entrees
-    if (tuple.length !== 2) {
-      return false
+  const validateWidthEquality = (updatedParams) => {
```
tannermpeterson, line 189: Doesn't look like this function is used anywhere so I think we can delete it

tannermpeterson (changes requested):

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -18,14 +17,24 @@ const Container = styled.div`
...
+  margin-bottom:4;
 `;
 
+const TwoParamContainer = styled.div`
+display: flex;
+flex-direction: column;
+height:100%;
+align-items:center;
+padding-top:2rem;
+`
```
luciipak, line 30: I would suggest installing prettier to fix some of these format issues in javascript files

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -143,65 +171,125 @@ export default function AnalysisParamForm({
...
+    if (newParams.prominenceFactorValleys !== undefined) {
+      validateProminenceFactorMax(updatedParams);
     }
-    if (newParams.widthFactor !== undefined) {
-      validateWidthFactor(updatedParams)
+    if (newParams.widthFactorPeaks !== undefined) {
+      validateWidthFactorMin(updatedParams)
+    }
+    if (newParams.widthFactorValleys !== undefined) {
+      validateWidthFactorMax(updatedParams)
```
luciipak, line 184: not necessarily a change, but you can just do `if (newParams.... )` or `if (!newParams...)`, you don't have to compare to undefined because undefined is a falsey value

frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -145,23 +144,40 @@ export default function UploadForm() {
...
+    if both are present then submit them as is
+    if none present then return null
+    if only peaks passed then make an array [peak,null]
+    if only valleys present then return array [null,valley]
+  */
+  const formatedAdvancedParams = (peaks, valleys) => {
+    if(peaks !== "" && valleys !== "") {return [peaks,valleys]}
+    else if(peaks !=="" && valleys === ""){return [peaks,null]}
+    else if(peaks === "" && valleys !== ""){return [null,valleys]}
+    else{return null}
```
luciipak, line 172: prettier

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
NikitaPoly: I had it installed but I had to also set it as default for javascript files

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -290,6 +299,25 @@ async def create_new_job(
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
 
+def _format_advanced_options(option: List[Union[int, float, None]], optionName):
+    if option is None:
+        return None
+    # if only peaks is passed return tupele(peaks,default value)
```
tannermpeterson, line 305: The typo is `tupele`

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -126,85 +152,121 @@ const AdditionalParamLabel = styled.span`
...
-    try {
-      parseFloat(tuple[0])
-      parseFloat(tuple[1])
-    } catch (e) {
-      return false
+  const isValidPositiveNumber = (value) => {
+    return +value > 0;
+  };
+
+  const validateProminenceFactorMin = (updatedParams) => {
```
tannermpeterson, line 193: These 4 validation functions are almost identical, you should  just make one function and just pass in the field you want to validate

tannermpeterson (changes requested):

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -290,6 +299,25 @@ async def create_new_job(
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
 
+def _format_advanced_options(option: List[Union[int, float, None]], optionName):
```
tannermpeterson, line 302: use snake_case instead of camelCase for python variable names (`option_name`)

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -190,80 +190,22 @@ export default function AnalysisParamForm({
...
       setParamErrors({
         ...paramErrors,
-        prominenceFactorPeaks: "",
+        paramName: "",
       });
     } else if (isValidPositiveNumber(newValue)) {
       setParamErrors({
         ...paramErrors,
-        prominenceFactorPeaks: "",
+        paramName: "",
```
tannermpeterson, line 203: since this a variable you'll need to `[paramName]: "",`

tannermpeterson (changes requested):

tannermpeterson (approved):

PR #176, opened 2022-08-18
added auto logout

NikitaPoly:


luciipak (changes requested):
Since we're removing the 401 responses from the individual routes getting called, we need to add handling in the service worker for that. Right now, you have it emitting a logout event when the tokens expire, we also want to emit an event if a 401 is returned at all

frontend/pulse3d/src/pages/login.js
```diff
@@ -66,7 +66,7 @@ export default function Login() {
               ? router.push("/new-user")
               : router.push("/uploads"); // routes to next page
           } else {
-            res.status === 401 || res.status === 422
+            res.status === 422
```
luciipak, line 69: somehow here, we will need to notify the user if the credentials are invalid. This would be the only request that might need 401 handling still. The other's shouldn't

frontend/pulse3d/src/pages/_app.js
```diff
@@ -48,11 +48,13 @@ function Pulse({ Component, pageProps }) {
       navigator.serviceWorker.addEventListener("message", ({ data }) => {
         // data returned is a boolean if auth tokens are present. Otherwise return user to login
         // might need auth check to include actual fetch request in SW to check token status if this becomes a problem
+        if (data.logout !== true) {
+          return;
+        }
```
luciipak, line 53: do we need `if (logout) setLoggedOutAlert(true);` ?

NikitaPoly: No I just tested it without it. I think this is left from my debug. I removed it.

luciipak (approved):

PR #177, opened 2022-08-19
Logout loop bug

NikitaPoly:


luciipak (approved):

frontend/pulse3d/src/serviceWorker.js
```diff
@@ -33,25 +33,25 @@ const tokens = {
 const setTokens = ({ access, refresh }) => {
   tokens.access = access.token;
   tokens.refresh = refresh.token;
-  setLogoutTimer();
```
tannermpeterson, line 36: I think we wanted to keep this line here, now it looks like if the tokens get refresh the logout timer doesn't get reset

NikitaPoly: I placed the function later on in the checks  and both ways see to work, but this is the simpler solution.

PR #178, opened 2022-08-20
add params for ci-terraform-apply workflowtemplate

jperson:
@luciipak Not sure if it needs sha or base-sha


luciipak:
Yeah pretty sure its sha

luciipak (approved):
Yeah just needs to be the same as the terraform plan step in the open pr trigger

PR #179, opened 2022-08-21
Uploads refactor

luciipak:


frontend/pulse3d/src/components/basicWidgets/DropDownWidget.js
```diff
@@ -77,17 +84,29 @@ export default function DropDownWidget({
...
+              <div>
+                <ListItem disabled={true}>{item}</ListItem>
+              </div>
+            </Tooltip>
+          ) : (
+            <ListItem key={idx} value={idx}>
+              {item}
+            </ListItem>
+          );
+        })}
```
luciipak, line 109: Hamed thought it would be better for the list items in 'Actions' to be disabled instead of the dropdown itself  so that users know what actions are available. Then each action has a tooltip that says they must select from the table below before they become available. No tooltip when actions are enabled.

jperson (approved):

PR #188, opened 2022-08-24
Admin uploads view

tannermpeterson:


frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -173,88 +173,48 @@ export default function AnalysisParamForm({
...
   const updateParams = (newParams) => {
-    setAllErrorMessagesBlank();
     const updatedParams = { ...analysisParams, ...newParams };
 
-    if (newParams.twitchWidths) {
+    if ("twitchWidths" in newParams) {
       validateTwitchWidths(updatedParams);
     }
-    if (newParams.startTime || newParams.endTime) {
+    if ("startTime" in newParams || "endTime" in newParams) {
```
luciipak, line 181: this is cool, I forgot you could do this with javascript

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -173,88 +173,48 @@ export default function AnalysisParamForm({
...
+    for (const paramName of [
+      "prominenceFactorPeaks",
+      "prominenceFactorValleys",
+      "widthFactorPeaks",
+      "widthFactorValleys",
+      "maxY",
+    ]) {
+      if (paramName in newParams) {
+        validatePositiveNumber(updatedParams, paramName, false);
+      }
```
luciipak, line 194: 👍🏽

frontend/pulse3d/src/serviceWorker.js
```diff
@@ -33,6 +31,7 @@ const tokens = {
 const setTokens = ({ access, refresh }) => {
   tokens.access = access.token;
   tokens.refresh = refresh.token;
+  setLogoutTimer();
```
luciipak, line 34: so this is where I was saying just clears the timeout and doesn't ensure that the `ClientSource.postMessage({ logout: true });` gets called inside the logout. Do you think that's something we should worry about or no?

tannermpeterson: I think `setLogoutTimer` does that, do you mean when the timer gets cleared?

luciipak: Yes. I'll try to investigate the order of requests made next time it happens, but clearing the timeout doesn't ensure that the client enforces the auto logout

luciipak: and the only time the logout message gets sent to _app.js is inside the timeout callback

luciipak: Also, my bad, I meant to addd this comment to:
`const clearTokens = () => {
  clearLogoutTimer();
};`

luciipak (approved):

PR #194, opened 2022-08-29
Fix build push + test

tannermpeterson:


core/db/curibio/alembic/versions/5b9e566ac996_.py
```diff
@@ -0,0 +1,24 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table("test", sa.Column("test", sa.String(12), primary_key=True))
+
+
+def downgrade():
+    op.drop_table("test")
```
luciipak, line 24: is this to test if it runs when merged?

luciipak (approved):

PR #203, opened 2022-08-30
fixed soft delete with account type user

luciipak:


core/lib/jobs/jobs/jobs.py
```diff
@@ -190,8 +190,11 @@ async def delete_jobs(*, con, account_type, account_id, job_ids):
 
     query_params = [account_id]
     places = _get_placeholders_str(len(job_ids), len(query_params) + 1)
-    if account_type == "users":
-        query = f"UPDATE jobs_result SET status='deleted' WHERE user_id=$1 AND job_id IN ({places})"
```
luciipak, line 194: user_id doesn't exist in jobs_result table

tannermpeterson: Yeah that's my bad

tannermpeterson (approved):

PR #210, opened 2022-09-07
added admin page to view and edit users

NikitaPoly:
-Added 3 new routes to users service
    1) fetching all users to display on the FE (/users)
    2) deleting one user at a time (/user-actions)
    3) deactivating one user at a time(/user-actions)

-Added a new route on the FE (/users-info)

### DIFF
--- a/deployments/apiv2/services/users/src/main.py
+++ b/deployments/apiv2/services/users/src/main.py
@@ -344,42 +344,32 @@ async def register(
 
 @app.get("/users")
 async def get_users(request: Request, token=Depends(ProtectedAny(scope=["users:admin"]))):
-    """Get all the information for all the users connected to the admin account that started this request.
+    """Get info for all the users under the given customer account.
+
     By default it will be sorted with all active users showing up first, then all the deactivated users
     """
     try:
         async with request.state.pgpool.acquire() as con:
-            list_of_users = []
             customer_id = uuid.UUID(hex=token["userid"])
-            active_rows = await con.fetch(
-                "SELECT name, email, created_at, last_login,suspended FROM users WHERE customer_id=$1 AND deleted_at IS NULL AND suspended='f'",
-                customer_id,
+
+            query = (
+                "SELECT name, email, created_at, last_login, suspended FROM users "
+                "WHERE customer_id=$1 AND deleted_at IS NULL "
+                "ORDER BY suspended"
             )
-            for row in active_rows:
-                list_of_users.append(
-                    {
-                        "name": row.get("name", ""),
-                        "email": row.get("email", ""),
-                        "date_created": row.get("created_at", ""),
-                        "last_loggedin": row.get("last_login"),
-                        "deactivated": row.get("suspended"),
-                    }
-                )
-            deactive_rows = await con.fetch(
-                "SELECT name, email, created_at, last_login,suspended FROM users WHERE customer_id=$1 AND deleted_at IS NULL AND suspended='t'",
-                customer_id,
-            )
-            for row in deactive_rows:
-                list_of_users.append(
-                    {
-                        "name": row.get("name", ""),
-                        "email": row.get("email", ""),
-                        "date_created": row.get("created_at", ""),
-                        "last_loggedin": row.get("last_login"),
-                        "deactivated": row.get("suspended"),
-                    }
-                )
-            return list_of_users
+            result = await con.fetch(query, customer_id)
+
+            return [
+                {
+                    "name": row.get("name", ""),
+                    "email": row.get("email", ""),
+                    "date_created": row.get("created_at", ""),
+                    "last_loggedin": row.get("last_login"),
+                    "deactivated": row.get("suspended"),
+                }
+                for row in result
+            ]
+
     except Exception as e:
         logger.exception(f"users GET: Unexpected error {repr(e)}")
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
--- a/deployments/apiv2/services/users/src/main.py
+++ b/deployments/apiv2/services/users/src/main.py
@@ -375,33 +375,29 @@ async def get_users(request: Request, token=Depends(ProtectedAny(scope=["users:a
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
 
-@app.delete("/user-actions/{user_email_to_delete}")
-async def delete_user(
-    request: Request, user_email_to_delete: EmailStr, token=Depends(ProtectedAny(scope=["users:admin"]))
-):
-    """Mark a users selected as deleted.When deleted the user will not be fetched when displaying table of users"""
-    try:
-        async with request.state.pgpool.acquire() as con:
-            update_query = "UPDATE users SET deleted_at=$1  WHERE email=$2"
-            await con.execute(update_query, datetime.datetime.now(), user_email_to_delete)
-    except Exception as e:
-        logger.exception(f"user-actions DELETE: Unexpected error {repr(e)}")
-        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
-
-
-@app.put("/user-actions/{user_email_to_deactivate}")
+@app.put("/users/{email_of_user_to_edit}")
 async def deactivate_user(
     request: Request,
     action_to_take: UserAction,
-    user_email_to_deactivate: EmailStr,
+    email_of_user_to_edit: EmailStr,
     token=Depends(ProtectedAny(scope=["users:admin"])),
 ):
-    """Deactivate user by setting suspended to true"""
+    """Edit a users information
+
+    The action to take on the user should be passed in the body of PUT request as action_type
+        -deactivate: set suspended field to true
+        -delete: set deleted_at field to current time
+    """
     try:
         async with request.state.pgpool.acquire() as con:
             if action_to_take.action_type == "deactivate":
                 update_query = "UPDATE users SET suspended='t' WHERE email=$1"
-                await con.execute(update_query, user_email_to_deactivate)
+                await con.execute(update_query, email_of_user_to_edit)
+
+            elif action_to_take.action_type == "delete":
+                update_query = "UPDATE users SET deleted_at=$1  WHERE email=$2"
+                await con.execute(update_query, datetime.datetime.now(), email_of_user_to_edit)
+
     except Exception as e:
         logger.exception(f"user-actions PUT: Unexpected error {repr(e)}")
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
--- a/deployments/apiv2/services/users/src/main.py
+++ b/deployments/apiv2/services/users/src/main.py
@@ -24,6 +24,7 @@ from models.users import (
     UserProfile,
     UserAction,
 )
+from pydantic import EmailStr
 from utils.db import AsyncpgPoolDep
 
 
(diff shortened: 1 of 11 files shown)


frontend/pulse3d/src/serviceWorker.js
```diff
@@ -92,7 +98,9 @@ const modifyRequest = async (req, url) => {
   const modifiedReq = new Request(getUrl(url), {
     headers,
     body:
-      req.method === "POST" ? JSON.stringify(await req.clone().json()) : null,
+      req.method === "POST" || req.method === "PUT"
```
tannermpeterson, line 101: Did you try `body: JSON.stringify(await req.clone().json()) || null`? I think Luci said that it works. It's a little cleaner than adding a bunch of checks on the method

deployments/apiv2/services/users/src/main.py
```diff
@@ -1,3 +1,4 @@
+import datetime
```
tannermpeterson, line 1: We also want to make sure that the new routes in this file have tests written for them

tannermpeterson (changes requested):

deployments/apiv2/services/users/src/main.py
```diff
@@ -327,3 +339,80 @@ async def register(
...
+
+
+@app.get("/users")
+async def get_users(request: Request, token=Depends(ProtectedAny(scope=["users:admin"]))):
+    """Get all the informatoin for all the users connected to the admin account that started this request.
+    By default it will be sorted with all active users showing up first, then all the deactivted users
+    """
+    try:
+        async with request.state.pgpool.acquire() as con:
+            listOfUsers = []
```
luciipak, line 351: snake case

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -0,0 +1,257 @@
...
+  };
+
+  const userActionSelection = async (option, name, email) => {
+    if (option === 1) {
+      setModalMessage(`Are you sure you would like to deactivate ${name}?`);
+      setActionToPreform("deactivate");
+      setEmailToDelete(email);
+    } else if (option === 0) {
+      setModalMessage(`Are you sure you would like to delete ${name}?`);
+      setActionToPreform("delete");
```
luciipak, line 65: setActionToPerform not Preform

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -0,0 +1,257 @@
...
+      setEmailToDelete(email);
+    }
+  };
+
+  const getAllUsers = async () => {
+    try {
+      const response = await fetch("https://curibio.com/users");
```
luciipak, line 72: I think this only needs to be https://curibio.com not https://curibio.com/users

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -0,0 +1,257 @@
...
+      if (response && response.status === 200) {
+        const usersJson = await response.json();
+        const userList = [];
+        for (const user of usersJson) {
+          userList.push(user);
+        }
```
luciipak, line 78: I don't think you need to create a new array and push the user in.. the usersJson looks to be an array already?

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -0,0 +1,257 @@
...
+    );
+  }, [currentPage]);
+
+  return (
+    <>
+      <ModalWidget
+        open={actionAlertVisible}
+        labels={[modalMessage]}
+        buttons={["yes", "no"]}
+        header={"attention"}
```
luciipak, line 121: I would capitalize the buttons and headers to match how the other modals are set up

frontend/pulse3d/src/serviceWorker.js
NikitaPoly: `body: JSON.stringify(await req.clone().json()) || null ` gives multiple errors

frontend/pulse3d/src/pages/users-info.js
NikitaPoly: Without the /users the user information can no be fetched.

deployments/apiv2/services/users/src/main.py
```diff
@@ -327,3 +340,78 @@ async def register(
...
+    except Exception as e:
+        logger.exception(f"users GET: Unexpected error {repr(e)}")
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+@app.delete("/user-actions/{user_email_to_delete}")
+async def delete_user(
+    request: Request, user_email_to_delete: EmailStr, token=Depends(ProtectedAny(scope=["users:admin"]))
+):
+    """Mark a users selected as deleted.When deleted the user will not be fetched when displaying table of users"""
```
tannermpeterson, line 392: The docstrings for our routes should focus more on what the route is going to do in the context of the BE. In this case, all we need to say is that it's going to mark the user with the given email as deleted in the DB. Someone or something could hit this route with no intention of displaying a table of users, so we probably shouldn't detail why the pulse3d webpage would hit this route

deployments/apiv2/services/users/src/main.py
```diff
@@ -327,3 +340,78 @@ async def register(
     except Exception as e:
         logger.exception(f"register: Unexpected error {repr(e)}")
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+@app.get("/users")
+async def get_users(request: Request, token=Depends(ProtectedAny(scope=["users:admin"]))):
+    """Get all the information for all the users connected to the admin account that started this request.
+    By default it will be sorted with all active users showing up first, then all the deactivated users
```
tannermpeterson, line 348: Is there any reason why we need to sort the users by their suspended status in this route? If sorting is needed in the FE, we could always handle it there

deployments/apiv2/services/users/src/main.py
```diff
@@ -327,3 +340,78 @@ async def register(
...
+                        "deactivated": row.get("suspended"),
+                    }
+                )
+            return list_of_users
+    except Exception as e:
+        logger.exception(f"users GET: Unexpected error {repr(e)}")
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+@app.delete("/user-actions/{user_email_to_delete}")
```
tannermpeterson, line 388: Is there any action other than setting the `deleted_at` column that might be performed here in the future? This route seems like it could just be `@app.delete("/users/{email}")`

deployments/apiv2/services/users/src/main.py
```diff
@@ -327,3 +340,78 @@ async def register(
...
+    try:
+        async with request.state.pgpool.acquire() as con:
+            update_query = "UPDATE users SET deleted_at=$1  WHERE email=$2"
+            await con.execute(update_query, datetime.datetime.now(), user_email_to_delete)
+    except Exception as e:
+        logger.exception(f"user-actions DELETE: Unexpected error {repr(e)}")
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+@app.put("/user-actions/{user_email_to_deactivate}")
```
tannermpeterson, line 402: Is there any reason we couldn't make this `@app.put("/users/{email}")`? I'm guessing that down the road we will want to make this a very general purpose route that can edit info about users, and to keep it more RESTful we could use this as the route name instead, as well as the other change I mentioned so that we have GET, PUT, and DELETE `/users/...`. Also if this route will end up being a general purpose route to edit users, we should also name the function something like `edit_user_info` and change the docstring

frontend/pulse3d/src/pages/_app.js
```diff
@@ -28,7 +28,7 @@ export const AuthContext = createContext();
 
 const availablePages = {
   user: ["/uploads", "/upload-form", "/account"],
-  admin: ["/uploads", "/new-user"],
+  admin: ["/uploads", "/new-user", "/users-info"],
```
tannermpeterson, line 31: We could probably just name this page `/users`. It hits a different server so it wouldn't collide with `/users` routes that we create in the BE

tannermpeterson (changes requested):

deployments/apiv2/services/users/tests/test_main.py
```diff
@@ -517,3 +517,128 @@ def test_logout__success(account_type, mocked_asyncpg_con):
...
+    )
+
+    assert response.status_code == 200
+
+    mocked_asyncpg_con.execute.assert_called_once_with(
+        f"UPDATE users SET deleted_at=$1  WHERE email=$2", datetime.now(), "testUser@curibio.com"
+    )
+
+
+def test_users__put_delete__no_email_given():
```
tannermpeterson, line 584: All the tests using `test_client.delete` need to be updated to use `put` since the delete method on this route got removed

deployments/apiv2/services/users/tests/test_main.py
```diff
@@ -517,3 +517,128 @@ def test_logout__success(account_type, mocked_asyncpg_con):
...
+    action_to_take = {"action_type": "delete"}
+
+    response = test_client.delete(
+        "/users", json=action_to_take, headers={"Authorization": f"Bearer {access_token}"}
+    )
+
+    assert response.status_code == 405
+
+
+def test_users__put_delete__invalid_token_scope_given():
```
tannermpeterson, line 596: The behavior being tested in this test is covered in `test_users__put__invalid_token_scope_given`, you can remove this test now

deployments/apiv2/services/users/tests/test_main.py
```diff
@@ -517,3 +517,128 @@ def test_logout__success(account_type, mocked_asyncpg_con):
...
+    action_to_take = {"action_type": "delete"}
+
+    response = test_client.delete(
+        "/users/test@email.com", json=action_to_take, headers={"Authorization": f"Bearer {access_token}"}
+    )
+
+    assert response.status_code == 405
+
+
+def test_users__put_deactivate__success(mocked_asyncpg_con):
```
tannermpeterson, line 608: nitpick, should make this test name `test_users__put__successful_deactivation`. The other test `test_users__put_delete__success` should use a similar naming pattern too

tannermpeterson (changes requested):

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -0,0 +1,254 @@
...
+  const [rows, setRows] = useState([]);
+  const [currentRows, setCurrentRows] = useState([]);
+  const [users, setUsers] = useState([]);
+  const [actionAlertVisible, setActionAlertVisible] = useState(false);
+  const [modalMessage, setModalMessage] = useState();
+  const [actionToPreform, setActionToPerform] = useState("");
+  const [confirm, setConfirm] = useState(false);
+  const [emailToEdit, setEmailToEdit] = useState("");
+
+  const deleteUser = async () => {
```
tannermpeterson, line 46: This and `deactivateUser` are almost identical, I think these could be one method that takes the action as a paremeter

deployments/apiv2/services/users/tests/test_main.py
```diff
@@ -517,3 +517,116 @@ def test_logout__success(account_type, mocked_asyncpg_con):
...
+    assert response.status_code == 200
+
+    mocked_asyncpg_con.execute.assert_called_once_with(
+        f"UPDATE users SET suspended='t' WHERE email=$1", "testUser@curibio.com"
+    )
+
+
+def test_users__put__no_action_given():
+    test_customer_id = uuid.uuid4()
+    expected_scope = ["users:admin"]
```
tannermpeterson, line 618: A lot of these tests don't use `expected_scope` in any assertions, you could just remove it and let `get_token` use the default scope for customer tokens

tannermpeterson (changes requested):

tannermpeterson (approved):

PR #222, opened 2022-09-15
initial waveform analysis

luciipak:


tannermpeterson:
Need to merge into RC branch instead of main

PR #230, opened 2022-09-21
Compatibility fixes

luciipak:


frontend/pulse3d/src/pages/uploads.js
```diff
@@ -144,14 +145,18 @@ export default function Uploads() {
...
+
             return {
               jobId: id,
               uploadId: upload_id,
               analyzedFile,
               datetime: formattedTime,
               status,
               analysisParams,
+              // version: pulse3dVersions[0], // tag with latest version for now, can't be before v0.25.1
+              version: "0.25.2",
```
luciipak, line 159: will change this in my next PR when i add dropdown.

frontend/pulse3d/src/.env.production
```diff
@@ -1,2 +1,2 @@
-NEXT_PUBLIC_PULSE3D_URL=https://pulse3d.curibio.com 
-NEXT_PUBLIC_USERS_URL=https://apiv2.curibio.com/users
+NEXT_PUBLIC_PULSE3D_URL=https://pulse3d.curibio-test.com 
+NEXT_PUBLIC_USERS_URL=https://apiv2.curibio-test.com/users
```
tannermpeterson, line 2: I was thinking we might want to create some new file, maybe `.env.testing` so this value doesn't get randomly changed. Or if this gets created automatically (I can't remember) maybe we just gitignore it

frontend/pulse3d/src/components/layouts/DashboardLayout.js
```diff
@@ -15,11 +16,15 @@ export const UploadsContext = createContext();
 export default function DashboardLayout({ children }) {
   const [uploads, setUploads] = useState([]);
   const [fetchUploads, setFetchUploads] = useState(false);
+  const [pulse3dVersions, setPulse3dVersions] = useState([]);
   const router = useRouter();
 
   useEffect(() => {
     if (router.pathname === "/uploads") {
       getUploads();
+      getPulse3dVersions();
```
tannermpeterson, line 25: The dashboard layout it always loaded right? If so, we could probably just hit this route and save the versions once, right when it gets loaded. Or if not, we could just check if `pulse3dVersions` and only hit the route if it's an empty list

frontend/pulse3d/src/components/uploads/TableRow.js
```diff
@@ -102,6 +102,8 @@ export default function Row({
   };
 
   const formatAnalysisParamColumn = (param, analyzedFile, value) => {
+    // Tanner (8/23/22): older analyses did not store the analysis params in the metadata of their DB entries,
```
tannermpeterson, line 105: I think this should be moved down to line 184 since that's the condition I had originally intended for this comment to explain

jobs/pulse3d/pulse3d-worker/src/main.py
```diff
@@ -112,7 +124,12 @@ async def process(con, item):
             try:
                 if not re_analysis:
                     logger.info("Writing time force data to parquet file for new upload")
-                    time_force_df = recordings[0].to_dataframe(tmpdir)
+                    if PULSE3D_VERSION == "0.24.6":
```
tannermpeterson, line 127: Can you add a TODO note here to use the semver package to compare if `PULSE3D_VERSION < "0.24.9"` instead

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -265,12 +265,17 @@ async def create_new_job(
             "twitch_widths",
             "start_time",
             "end_time",
-            "peaks_valleys",
         ]
         # TODO could make this if condition `details.version <= "0.25.0"` using the semver package
         if details.version != "0.24.6":
             # max_y param was added in 0.25.0
             params.append("max_y")
+        if details.version in (
```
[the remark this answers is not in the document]
tannermpeterson: Might also be good to add some tests for this condition, but if we just want to get this out ASAP you could just add a TODO note to add tests for it later

frontend/pulse3d/src/.env.production
luciipak: yeah lets git ignore them.. there is some issues with exporting in next because the "test env" is considered when you run `npm run dev` so even though we're testing here, it's still considered production to nextjs.

luciipak: That's why we're dynamically changing them each time,  but we could probably find a way to use diff files in the future.

frontend/pulse3d/src/components/layouts/DashboardLayout.js
luciipak: it refreshes though when the router changes at the moment which I think happens between the uploads page and the uploads-form but I could be wrong..

tannermpeterson (approved):

PR #233, opened 2022-09-22
Y axis normalization option

NikitaPoly:
Added option to disable y axis normalization.
When y axis normalization  is disabled, then max_y is also disabled


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -259,6 +259,7 @@ async def create_new_job(
         logger.info(f"Creating pulse3d job for upload {details.upload_id} with user ID: {user_id}")
 
         params = [
+            "normalize_y_axis",
```
luciipak, line 262: Depending on which version of Pulse3D this param gets added, you'll need to conditionally add it in. See lines 271-281

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -269,7 +269,9 @@ async def create_new_job(
         # TODO could make this if condition `details.version <= "0.25.0"` using the semver package
         if details.version != "0.24.6":
             # max_y param was added in 0.25.0
+            # normalize_y_axis added in 0.25.4
             params.append("max_y")
+            params.append("normalize_y_axis")
```
luciipak, line 274: this means it will still be added with versions 0.25.1 and 0.25.2

luciipak:
We might want to add 0.25.4 to the DB and bump the pulse3d version in pulse3d-worker/pulse3d service as well

deployments/pulse3d/services/pulse3d/src/main.py
tannermpeterson: Yeah we really need to start using the semver package to manage all this version conflicts

tannermpeterson:
Yeah we have to do that. It might make sense to do start using DB migrations to bump the pulse3d version so we don't have to manually add things to main when the RC branches get merged in

deployments/pulse3d/services/pulse3d/src/main.py
NikitaPoly: Would adding the _string_version_to_numbers be a good solution for now?

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -250,6 +251,10 @@ async def _get_jobs(con, token, job_ids):
     return await get_jobs(con=con, account_type=account_type, account_id=account_id, job_ids=job_ids)
 
 
+def _string_version_to_numbers(version_string: string) -> List[int]:
```
tannermpeterson, line 254: @NikitaPoly there is an existing package that handles comparing semvers https://pypi.org/project/semver/, we should use it instead of writing our own code to parse strings into ints

deployments/pulse3d/services/pulse3d/src/main.py
tannermpeterson: The semver package is pretty simple to use, so we should just add it now. You can see an example of using it here https://github.com/CuriBio/pulse3d/blob/2858dbae54772a03ac3543e8456ff6b32c1d4142/src/pulse3D/plate_recording.py#L234

tannermpeterson (changes requested):

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -333,6 +343,39 @@ export default function AnalysisParamForm({
             />
           </DropDownContainer>
         </ParamContainer>
+
+        <ParamContainer style={{ width: "33%", marginTop: "2%" }}>
```
tannermpeterson, line 347: This should only be rendered if the selected pulse3d version supports this param

tannermpeterson (changes requested):

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
NikitaPoly: Should I use a similar JS library for checking versions?

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -266,14 +267,13 @@ async def create_new_job(
...
-        # TODO could make this if condition `details.version <= "0.25.0"` using the semver package
-        if details.version != "0.24.6":
+        if details.version.split(".") >= VersionInfo.parse("0.25.0"):
             # max_y param was added in 0.25.0
             params.append("max_y")
-        if details.version not in (
-            "0.24.6",
-            "0.25.1",
-        ):
+        if details.version.split(".") >= VersionInfo.parse("0.25.4"):
```
tannermpeterson, line 273: 👍

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -333,7 +344,42 @@ export default function AnalysisParamForm({
             />
           </DropDownContainer>
         </ParamContainer>
-        {inputVals.selectedPulse3dVersion !== "0.24.6" && (
+        {semverGte(inputVals.selectedPulse3dVersion, "0.25.4") && (
```
tannermpeterson, line 347: 👍

deployments/pulse3d/services/pulse3d/src/main.py
tannermpeterson: The last thing you need to do is bump the pulse3d version in the worker to this version and add the scaledjob for it in the manifests

tannermpeterson (changes requested):

core/db/curibio/alembic/versions/c1ae9bb9fa14_pulse3_0_25_4.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'c1ae9bb9fa14'
+down_revision = '871f6d005d86'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    pass
```
tannermpeterson, line 20: You'll need to implement `upgrade` and `downgrade`. You can take a look at some of the other migrations to see how they work

tannermpeterson (changes requested):

core/db/curibio/alembic/versions/c1ae9bb9fa14_pulse3_0_25_4.py
```diff
@@ -0,0 +1,24 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    query = "INSERT INTO pulse3d_versions (version) VALUES ('0.25.4')"
+    op.execute(query)
+
+
+def downgrade():
```
tannermpeterson, line 23: You'll need to define the downgrade too. In this case it should just remove the row that `upgrade` adds

tannermpeterson (changes requested):

core/db/curibio/alembic/versions/c1ae9bb9fa14_pulse3_0_25_4.py
```diff
@@ -0,0 +1,25 @@
...
+    query = "DELETE FROM pulse3d_versions WHERE versions != '0.25.4'"
```
tannermpeterson, line 24: This will delete every version from the table except 0.25.4, we want to delete only 0.25.4. Also the column name should be `version` not `versions`

tannermpeterson (changes requested):

tannermpeterson (approved):

PR #241, opened 2022-09-29
deactivated users cannot login

luciipak:


tannermpeterson (commented):
I think you might need to update the unit tests

jperson (approved):

PR #247, opened 2022-10-06
Table sorting

NikitaPoly:
Added pagination and filtering to uploads and user table using https://github.com/jbetancur/react-data-table-component.


frontend/pulse3d/src/pages/uploads.js
```diff
@@ -175,12 +307,13 @@ export default function Uploads() {
   useEffect(() => {
     getAllJobs();
     // start 10 second interval
+    /*
```
luciipak, line 310: I think  we want to keep this uploadsInterval.. I'm not sure why the initial state is also commented out. it's being used in the RC10132022 branch

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -192,7 +325,9 @@ export default function Uploads() {
           .filter(({ uploadId }) => uploadId === id)
           .sort((a, b) => new Date(b.datetime) - new Date(a.datetime));
 
-        const lastAnalyzed = uploadJobs[0] ? uploadJobs[0].datetime : formattedTime;
+        const lastAnalyzed = uploadJobs[0]
+          ? uploadJobs[0].datetime
+          : formattedTime;
```
luciipak, line 330: Seems like your prettier config is setting things back to a line length of 80.. might check to see if we can move everything back to line length of 110 because we'll keep having to change the line length back and forth in PRs.

frontend/pulse3d/src/components/table/UploadsSubTable.js
```diff
@@ -0,0 +1,87 @@
...
+          </div>
+        );
+      }
+    });
+
+    return (
+      <SubContainer key={Math.random()}>
+        <input
+          type="checkbox"
+          checked={isChecked[idx]}
```
luciipak, line 49: There is a checkbox widget that you can customize the selection color, but not crucial

frontend/pulse3d/src/pages/uploads.js
luciipak: Could you actually merge the RC10132022 branch into this one before merging? it's missing a few changes and just want to make sure they don't get overridden

NikitaPoly: I think I already merged it. I can fix this issue manually As well. Il make sure to keep the updates for job status

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -104,6 +175,61 @@ export default function Uploads() {
...
+      // don't call get jobs if downloading ro deleting in progress because it backs up server
+      if (!["downloading", "deleting"].includes(modalState) && updateData) {
+        await getAllJobs();
+        toggleUpdateData(!updateData);
+      }
+    }, [1e4]);
+  }, [updateData]);
+  useEffect(() => {
+    console.log("checked time out in the effect is : " + checkedJobs);
+  }, [checkedJobs]);
```
luciipak, line 194: Is this useEffect necessary?

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -104,6 +175,61 @@ export default function Uploads() {
...
+
+  useEffect(() => {
+    if (displayRows.length > 0) {
+      setTimeout(() => {
+        setPending(false);
+      }, 5000);
+    }
+  }, [displayRows]);
+  const tofilterField =
```
luciipak, line 203: toFilterField

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -116,21 +242,21 @@ export default function Uploads() {
...
-
-        jobs = jobs.map(({ id, upload_id, created_at, object_key, status, meta }) => {
+        const newJobs = jobs.map(({ id, upload_id, created_at, object_key, status, meta }) => {
           const analyzedFile = object_key ? object_key.split("/")[object_key.split("/").length - 1] : "";
           const formattedTime = formatDateTime(created_at);
           const parsedMeta = JSON.parse(meta);
           const analysisParams = parsedMeta.analysis_params;
-
+          const isChecked = checkedJobs.includes(id);
+          console.log(checkedJobs.includes(id) + " : maybe");
```
luciipak, line 259: remove console.log

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -175,12 +301,15 @@ export default function Uploads() {
...
+    setTimeout(() => {
       // don't call get jobs if downloading ro deleting in progress because it backs up server
-      if (!["downloading", "deleting"].includes(modalState)) getAllJobs();
+      if (!["downloading", "deleting"].includes(modalState)) {
+        toggleUpdateData(!updateData);
+      } //getAllJobs();
     }, [1e4]);
     //clear interval when switching pages
-    return () => clearInterval(uploadsInterval);
+    //return () => clearInterval(uploadsInterval);
```
luciipak, line 312: there are two of these set timeouts... I'd keep this one and remove the other one

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -193,7 +322,7 @@ export default function Uploads() {
           .sort((a, b) => new Date(b.datetime) - new Date(a.datetime));
 
         const lastAnalyzed = uploadJobs[0] ? uploadJobs[0].datetime : formattedTime;
-
+        console.log();
```
luciipak, line 325: remove console.log

frontend/pulse3d/src/components/table/FilterHeader.js
```diff
@@ -0,0 +1,58 @@
...
+
+  //when input changes update the filterstring
+  useEffect(() => {
+    props.setFilterString(input);
+  }, [input]);
+
+  const searchFields = props.columns.map((column, idx) => {
+    if (!props.loading) {
+      return column === "" ? (
+        <input
```
tannermpeterson, line 31: Is there a way to just return one input component and just conditionally choose which props to pass to it? I think that would make it a little easier to read and tell what the difference between column === "" and the opposite are. There's also a lot that's the same here like onChange, type, and key, so it would be nice to not have that duplicated

frontend/pulse3d/src/components/table/UploadsSubTable.js
```diff
@@ -0,0 +1,71 @@
...
+
+export default memo(function UploadsSubTable(props) {
+  const [isChecked, setIsChecked] = useState(props.jobs.map((job) => job.checked));
+  const rows = props.jobs.map((job, idx) => {
+    let paramsString = [];
+    Object.keys(job.analysisParams).forEach((param) => {
+      if (job.analysisParams[param] !== null) {
+        if (param === "peaks_valleys") {
+          paramsString.push(<div key={Math.random()}>{`${param} : user set`}</div>);
+          return;
```
tannermpeterson, line 31: I think you could just use if/else here instead of returning, unless I'm missing something. Or better yet something like this
```
const paramVal = job.analysisParams[param] ? "user set" : job.analysisParams[param]
paramsString.push(<div key={Math.random()}>{`${param} : ${paramVal}`}</div>);
```

frontend/pulse3d/src/components/table/UploadsSubTable.js
```diff
@@ -0,0 +1,71 @@
...
+      }
+    });
+
+    return (
+      <SubContainer key={Math.random()}>
+        <input
+          type="checkbox"
+          checked={isChecked[idx]}
+          onChange={() => {
+            if (!isChecked[idx]) {
```
tannermpeterson, line 43: The only difference between the if/else branches here is `action: "add"/"remove"`. You can simplify this a lot by just doing
```
props.setJobToEdit({ id: job.jobId, action: isChecked[idx] ? "remove" : "add" });
setIsChecked(isChecked.map((checked, index) => (idx === index ? !checked : checked)));
```

tannermpeterson (changes requested):

frontend/pulse3d/src/components/table/UsersActionsSelector.js
```diff
@@ -0,0 +1,47 @@
...
+    <>
+      <Container>
+        <button
+          onClick={() => {
+            buttonAction("delete");
+          }}
+        >
+          Delete
+        </button>
+        {!props.data.suspended ? (
```
tannermpeterson, line 35: I think for conditionally rendering things you can just do
```
!props.data.suspended && (
  ... // component
)
```
Instead of using the ternary operator

frontend/pulse3d/src/components/table/UploadsSubTable.js
luciipak: I also thought it might be more user-friendly to remove the snake casing from the param names shown in this table column

luciipak: I think Tanner was saying this could be simplified

tannermpeterson: Yeah, with the snake case change it could be this now
```
const paramVal = param === "peaks_valleys" ? "user set" : job.analysisParams[param];
paramsString.push(<div key={Math.random()}>{`${param.replace("_", " "): ${paramVal}`}</div>);
```

Also why is `Math.random()` used for the key here? If you set it to string containing the job ID + param and it should always be unique. That would also make it easier to find the element in the HTML if that ever needed to be done

tannermpeterson (approved):

PR #252, opened 2022-10-12
Tag pulse versions with testing

luciipak:


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -530,7 +530,7 @@ async def get_versions(request: Request):
                 "SELECT * FROM pulse3d_versions WHERE state != 'deprecated' ORDER BY created_at"
             )
 
-        return [row["version"] for row in rows]
+        return [{"version": row["version"], "state": row["state"]} for row in rows]
```
tannermpeterson, line 533: You could update the the query to `SELECT version, state FROM ...` since the other columns aren't used, and then just do `[dict(row) for row in rows]`, but not a big deal

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -154,7 +154,13 @@ export default function AnalysisParamForm({
 
   const pulse3dVersionGte = (version) => {
     const { selectedPulse3dVersion } = inputVals;
-    return selectedPulse3dVersion && semverGte(selectedPulse3dVersion, version);
+
+    // this currently gets called on render when selectedPulse3dVersion is undefined
+    const versionNameOnly = selectedPulse3dVersion
```
tannermpeterson, line 159: Is there anyway the "testing" label could be just applied to the display names in the dropdown to allow the versions to be stored just as semvers? That would make it so you don't have to split the tag off everywhere you need to use the actual version

tannermpeterson (approved):

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
luciipak: Yeah I was thinking about a better way to handle it.. but it will need to be handled either way.. if we just add it to the dropdown options, we'd need to store an  array of just semver versions and then an array with the states, then constantly switch between the two during selections. I'll give it a go

deployments/pulse3d/services/pulse3d/Makefile
```diff
@@ -12,13 +12,13 @@ buildx:
 	docker buildx build -t pulse3d_api . --platform linux/amd64 --load -f ${repo_root}/deployments/pulse3d/services/pulse3d/Dockerfile
 
 tag:
-	docker tag pulse3d_api:latest $(ecr_repo):latest
+	docker tag pulse3d_api:latest $(ecr_repo):0.1.0
```
tannermpeterson, line 15: Do we need to do this for any of the other Makefiles of any other svcs?

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -305,7 +307,12 @@ export default function AnalysisParamForm({
           </Label>
           <DropDownContainer>
             <DropDownWidget
-              options={pulse3dVersions}
+              options={pulse3dVersions.map((version) => {
```
tannermpeterson, line 310: 👍

tannermpeterson (approved):

PR #254, opened 2022-10-13
when uploads checkbox is checked then jobs checkboxes also become checked

NikitaPoly:
Also changed alignment of table items to left


frontend/pulse3d/src/pages/uploads.js
```diff
@@ -640,6 +650,20 @@ export default function Uploads() {
...
+                        for (let i = 0; i < jobs.length; i++) {
+                          if (jobs[i].jobId === job.jobId) {
+                            jobs[i].checked = true;
+                            setCheckedJobs([...checkedJobs, jobs[i].jobId]);
+                          }
+                        }
+                      }
+                    });
+                  });
+                }
```
luciipak, line 666: Say a user selects an upload to check all the corresponding jobs, does this include a way to uncheck the upload row if a user unchecks one of the selected jobs? and then a way to uncheck all jobs if a user unchecks an upload row?

NikitaPoly: No not yet. I will add this

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -14,30 +14,35 @@ let adminColumns = [
     name: "File Owner",
     center: true,
     sortable: true,
+    center: false,
```
tannermpeterson, line 17: `center: true,` is two lines above this, couldn't you just change that line instead of adding all these new ones?

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -10,30 +10,35 @@ const columns = [
     name: "Status",
     center: true,
     sortable: true,
+    center: false,
```
tannermpeterson, line 13: ditto

PR #255, opened 2022-10-14
RC10132022

tannermpeterson:


core/db/curibio/alembic/versions/705b8a7903c8_pulse3d_0_27_0.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "705b8a7903c8"
+down_revision = "064b7f68d6c4"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("INSERT INTO pulse3d_versions (version) VALUES ('0.27.0')")
```
luciipak, line 19: might need to set the state to 'external' to show up in the prod cluster dropdown

tannermpeterson: We can just manually update it for this one so we can get it merged quicker

jperson: let's merge as is and if it needs it we can do a quick PR with another upgrade that sets the state

jperson (approved):

PR #264, opened 2022-10-18
Table checkbox

NikitaPoly:


frontend/pulse3d/src/pages/uploads.js
```diff
@@ -14,25 +14,30 @@ const uploadTableColumns = [
...
   },
   {
     name: "Created Date",
     admin: false,
+    center: false,
     selector: (row) => row.createdAt,
   },
   {
     name: "Last Analyzed",
+    center: false,
```
tannermpeterson, line 40: If these should all be `false` now, you can just change the value on line 48

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -9,22 +9,38 @@ import CircularSpinner from "@/components/basicWidgets/CircularSpinner";
...
     name: "Date Created",
+
+    center: false,
+    sortable: true,
+
     selector: (row) => formatDateTime(row.created_at),
   },
   {
     name: "Last Loggedin",
+    center: false,
```
tannermpeterson, line 42: these values are defined on line 49 now

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -273,36 +278,31 @@ export default function Uploads() {
   useEffect(() => {
     getAllJobs();
     // start 10 second interval
-    const uploadsInterval = setInterval(() => getAllJobs(), [1e4]);
```
luciipak, line 276: How come we want to remove this?

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -273,36 +278,31 @@ export default function Uploads() {
   useEffect(() => {
     getAllJobs();
     // start 10 second interval
```
luciipak, line 280: can you remove this comment

frontend/pulse3d/src/pages/uploads.js
tannermpeterson, line 276: If this is getting removed in favor of another method of polling the job status, then I don't think it is working currently. If there is new code to do this, it needs to be fixed, or this should just be added back

tannermpeterson (changes requested):

PR #267, opened 2022-10-20
added 0.27.1 pulse3d

NikitaPoly:
Added version 0.27.1 of pulse3D

### DIFF
--- a/core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
+++ b/core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
@@ -18,9 +18,9 @@ depends_on = None
 
 def upgrade():
     op.execute("INSERT INTO pulse3d_versions (version, state) VALUES ('0.27.1', 'external')")
-    op.execute("DELETE FROM pulse3d_versions WHERE version='0.27.0'")
+    op.execute("UPDATE pulse3d_versions SET state='deprecated' WHERE version='0.27.0'")
 
 
 def downgrade():
     op.execute("DELETE FROM pulse3d_versions WHERE version='0.27.1' AND state='external'")
-    op.execute("INSERT INTO pulse3d_versions (version) VALUES ('0.27.0')")
+    op.execute("UPDATE pulse3d_versions SET state=NULL WHERE version='0.27.0'")
--- a/core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
+++ b/core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
@@ -22,5 +22,5 @@ def upgrade():
 
 
 def downgrade():
-    op.execute("DELETE FROM pulse3d_versions WHERE version='0.27.1' AND state='external'")
+    op.execute("DELETE FROM pulse3d_versions WHERE version='0.27.1'")
     op.execute("UPDATE pulse3d_versions SET state='external' WHERE version='0.27.0'")
--- a/core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
+++ b/core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
@@ -17,7 +17,7 @@ depends_on = None
 
 
 def upgrade():
-    op.execute("INSERT INTO pulse3d_versions (version) VALUES ('0.27.1')")
+    op.execute("INSERT INTO pulse3d_versions (version, state) VALUES ('0.27.1', 'external')")
 
 
 def downgrade():
--- a/core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
+++ b/core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
@@ -1,0 +1,24 @@
+"""pulse3D 0.27.1
+
+Revision ID: 5d220341bf80
+Revises: 705b8a7903c8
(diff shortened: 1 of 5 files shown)


core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "5d220341bf80"
+down_revision = "705b8a7903c8"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("INSERT INTO pulse3d_versions (version) VALUES ('0.27.1')")
```
luciipak, line 20: will you add "INSERT INTO pulse3d_versions (version, state) VALUES ('0.27.1', 'external')"

NikitaPoly: I will add that, what does the external specify?

NikitaPoly: Did you mean replace the query ?

luciipak: We've changed it so that only versions marked 'external' will be shown to customers in production so that they don't get cluttered with all the versions we release during a sprint cycle.. this is what allows us to show which mark versions as 'testing' in the test cluster and allow us to use all the versions before getting released

jperson: can we also delete version 0.27.0, since it has a bug and people shouldn't use it

jobs/manifests/overlays/prod/kustomization.yaml
```diff
@@ -3,45 +3,58 @@ kind: Kustomization
...
+    patch: |-
+      - op: replace
+        path: /spec/jobTargetRef/template/spec/containers/0/image
+        value: "245339368379.dkr.ecr.us-east-2.amazonaws.com/pulse3d-worker:0.27.0"
+      - op: replace
+        path: /spec/jobTargetRef/template/spec/containers/0/env/1/value
+        value: prod-pulse3d-uploads
+      - op: replace
+        path: /spec/jobTargetRef/template/spec/containers/0/env/2/value
+        value: prod-mantarray-logs
```
luciipak, line 47: since we're deleting 0.27.0, we can delete all related jobs

jobs/manifests/overlays/prod/kustomization.yaml
```diff
@@ -3,45 +3,45 @@ kind: Kustomization
 
 namespace: pulse3d
 resources:
-- ../../base
+  - ../../base
```
tannermpeterson, line 6: Can you try running pre-commit to remove all this whitespace change

luciipak:
@tannermpeterson do we need to bump the version in the requirements.txt for the job to 0.27.1?

core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = "5d220341bf80"
+down_revision = "705b8a7903c8"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("INSERT INTO pulse3d_versions (version, state) VALUES ('0.27.1', 'external')")
+    op.execute("DELETE FROM pulse3d_versions WHERE version='0.27.0'")
```
tannermpeterson, line 21: We want to mark this as deprecated instead of deleting

tannermpeterson (changes requested):

tannermpeterson:
@luciipak yes

core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def upgrade():
+    op.execute("INSERT INTO pulse3d_versions (version, state) VALUES ('0.27.1', 'external')")
+    op.execute("UPDATE pulse3d_versions SET state='deprecated' WHERE version='0.27.0'")
+
+
+def downgrade():
+    op.execute("DELETE FROM pulse3d_versions WHERE version='0.27.1' AND state='external'")
+    op.execute("UPDATE pulse3d_versions SET state=NULL WHERE version='0.27.0'")
```
tannermpeterson, line 26: external not null

tannermpeterson (changes requested):

core/db/curibio/alembic/versions/5d220341bf80_pulse3d_0_27_1.py
```diff
@@ -0,0 +1,26 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.execute("INSERT INTO pulse3d_versions (version, state) VALUES ('0.27.1', 'external')")
+    op.execute("UPDATE pulse3d_versions SET state='deprecated' WHERE version='0.27.0'")
+
+
+def downgrade():
+    op.execute("DELETE FROM pulse3d_versions WHERE version='0.27.1' AND state='external'")
```
tannermpeterson, line 25: Don't need `AND state='external'`

Also still need to bump the pulse3d version in the pulse3d job

tannermpeterson (changes requested):

tannermpeterson (approved):

PR #268, opened 2022-10-21
Uploads checkbox

NikitaPoly:
Added checkbox logic
Filtering is now none case sensitive
Fixed styling of tables
Fixed interval issue (now clears when page is changed)
checked that uploads and jobs actions work
checked that jobs and uploads get cleared after action


deployments/apiv2/services/users/src/core/config.py
```diff
@@ -1,7 +1,7 @@
 from starlette.config import Config
 from starlette.datastructures import Secret
 
-config = Config(".env")
+config = Config("~/Documents/gpg/.env_localhost")
```
luciipak, line 4: Can you switch these back?

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -156,6 +144,46 @@ export default function Uploads() {
...
+    },
+    {
+      name: "Last Analyzed",
+      center: false,
+      admin: false,
+      selector: (row) => row.lastAnalyzed,
+    },
+  ];
   useEffect(() => {
     setTimeout(async () => {
```
luciipak, line 188: did you confirm that this setTimeout works and updates the job statuses without having to refresh the page?

frontend/pulse3d/src/serviceWorker.js
```diff
@@ -248,7 +248,7 @@ self.onmessage = ({ data, source }) => {
     });
   } else if (data.msgType === "stayAlive") {
     // TODO should have this do something else so that there isn't a log msg produced every 20 seconds
-    console.log("Staying alive");
+    // console.log("Staying alive");
```
luciipak, line 251: can you add this back in?

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -7,34 +7,49 @@ import FilterHeader from "@/components/table/FilterHeader";
...
+
+    center: false,
+    sortable: true,
+
     selector: (row) => formatDateTime(row.created_at),
   },
   {
     name: "Last Loggedin",
+    center: false,
+    sortable: true,
```
luciipak, line 37: you could probably add these to the `columnProperties` on line 49

frontend/pulse3d/src/serviceWorker.js
tannermpeterson: When you do remove things like this, you should also remember to remove the comments associated with them. This is mentioned in one of the comments above as well

deployments/pulse3d/services/pulse3d/src/core/config.py
```diff
@@ -1,7 +1,8 @@
 from starlette.config import Config
 from starlette.datastructures import Secret
 
-config = Config(".env")
+# config = Config(".env")
+config = Config("~/Documents/gpg/.env_localhost")
```
tannermpeterson, line 5: ditto to Luci's note

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -156,6 +144,46 @@ export default function Uploads() {
...
+      center: false,
+      admin: false,
+      selector: (row) => (
+        <Checkbox id={row.id} checked={checkedUploads.includes(row.id)} onChange={handleCheckedUploads} />
+      ),
+    },
+    {
+      name: "File Owner",
+      admin: true,
+      center: false,
```
tannermpeterson, line 159: This value is defined in `columnProperties`, you can remove it here and all the other objs in this array

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -236,8 +260,8 @@ export default function Uploads() {
             datetime: formattedTime,
             status,
             analysisParams,
-            // version: pulse3dVersions[0], // tag with latest version for now, can't be before v0.25.1
-            version: "0.25.2",
+            version: pulse3dVersions[0], // tag with latest version for now, can't be before v0.25.1
+            // version: "0.25.2",
```
tannermpeterson, line 264: Remove comment

tannermpeterson (changes requested):

deployments/pulse3d/services/pulse3d/src/core/config.py
luciipak: Some of these were my fault

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -139,7 +127,7 @@ const modalObjs = {
```
luciipak, line 127: and I just want to double check that all these modals are still working as expected? ensuring no one performs actions on pending jobs and that you can't download or perform interactive analysis on failed jobs

frontend/pulse3d/src/serviceWorker.js
luciipak: I did this one, I thought only part of the branch would be used. My bad

frontend/pulse3d/src/pages/uploads.js
NikitaPoly: I have confirmed that it does send the request.

frontend/pulse3d/src/pages/uploads.js
NikitaPoly: I just tested these features and they seem to work.

luciipak:
Okay not to be picky, I like that the filter inputs are now more centered to the column they relate to, but I think they should all be the same width.
Screen Shot 2022-10-21 at 3 47 57 PM
Screen Shot 2022-10-21 at 3 48 23 PM

frontend/pulse3d/src/components/layouts/DashboardLayout.js
```diff
@@ -74,7 +74,13 @@ export default function DashboardLayout({ children }) {
       value={{ uploads, setUploads, setFetchUploads, pulse3dVersions, metaPulse3dVersions }}
     >
       <Container>
-        <ControlPanel />
+        <ControlPanel
+          clearTimers={() => {
```
tannermpeterson, line 78: I don't think you need this check here, the same one is done when `ControlPanel` actually tries to to call the function

frontend/pulse3d/src/components/layouts/DashboardLayout.js
```diff
@@ -74,7 +74,13 @@ export default function DashboardLayout({ children }) {
...
+            if (clearTimers) {
```
tannermpeterson, line 79: I don't think you need this check here, the same one is done when `ControlPanel` actually tries to to call the function

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -274,16 +274,19 @@ export default function Uploads() {
       return datetime;
     }
   };
+  useEffect(() => {}, []);
```
tannermpeterson, line 277: Is this necessary?

frontend/pulse3d/src/components/layouts/DashboardLayout.js
tannermpeterson, line 79: I think it's better to leave the check in ControlPanel and remove it here since it's actually getting called in `ControlPanel`. If another component were to instantiate a `ControlPanel` it would also need to add this check, vs. if we just put the check in `ControlPanel` then it will only need to be done once no matter how many other components use `ControlPanel`

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -272,37 +277,35 @@ export default function Uploads() {
...
+        statusUpdateInterval = setInterval(async () => {
+          if (!["downloading", "deleting"].includes(modalState)) {
+            await getAllJobs();
+          }
+        }, [1e4]);
+      }
     }
-
-    //clear interval when switching pages
-    return () => clearInterval(uploadsInterval);
```
luciipak, line 283: why don't you add this back in instead of passing clearInterval to the layout? the return clears the interval when you switch pages which I think is what you're doing by adding it to the onClick of the accordion tabs

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -272,37 +277,35 @@ export default function Uploads() {
...
     getAllJobs();
-    // start 10 second interval
-    const uploadsInterval = setInterval(() => getAllJobs(), [1e4]);
     // don't call get jobs if downloading or deleting in progress because it backs up server
-    if (!["downloading", "deleting"].includes(modalState)) {
-      toggleUpdateData(!updateData);
+    if (!["downloading", "deleting"].includes(modalState) && uploads.length > 0) {
+      if (!statusUpdateInterval) {
+        statusUpdateInterval = setInterval(async () => {
+          if (!["downloading", "deleting"].includes(modalState)) {
```
luciipak, line 284: also you have the same conditional twice

tannermpeterson (approved):

PR #273, opened 2022-10-24
added file names to error message

NikitaPoly:
Error message  in /uploads-form


luciipak (commented):
What was the issue and how does this fix it? was reset getting called before the error modal gets opened?

NikitaPoly:
the file names were being overwritten by a default error message. This is after the update.
Capture

luciipak:
Got it yeah, could you reset the modal state in the `handleClose` function when user closes this? it will need to get reset at some point so it does't keep pushing filenames.

luciipak (approved):

PR #276, opened 2022-10-25
fixed filters

NikitaPoly:
Uploads filter won't reset on each load.
Fixed date filtering.

### DIFF
--- a/frontend/pulse3d/src/pages/users-info.js
+++ b/frontend/pulse3d/src/pages/users-info.js
@@ -20,7 +20,7 @@ const columns = [
     selector: (row) => formatDateTime(row.created_at),
   },
   {
-    name: "Last Loggedin",
+    name: "Last Log In",
     selector: (row) => formatDateTime(row.last_login),
   },
   {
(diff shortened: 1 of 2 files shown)


frontend/pulse3d/src/pages/users-info.js
```diff
@@ -146,9 +146,13 @@ export default function UserInfo() {
   useEffect(() => {
     const newList = usersData.filter((user) => {
       //if the column containes date data
-      return formatDateTime(user[toUserField[filtercolumn]])
-        .toLocaleLowerCase()
-        .includes(filterString.toLocaleLowerCase());
+      if (["Date Created", "Last Loggedin"].includes(filtercolumn)) {
```
luciipak, line 149: you just removed "Date Created", "Last Loggedin" from the column names above? is that right?

NikitaPoly: @luciipak  No that was in the Uploads table. This is in the User table

tannermpeterson: This is on the users table, I think it's still there. We need to change `Loggedin` to something else since it is customer facing, I think we should make it `Last Login`

luciipak: Got it, my bad on that

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -20,7 +20,7 @@ const columns = [
     selector: (row) => formatDateTime(row.created_at),
   },
   {
-    name: "Last Loggedin",
+    name: "Last Log In",
```
luciipak, line 23: Login is one word

luciipak (approved):

PR #283, opened 2022-10-28
Cloud usage restrictions

luciipak:


core/lib/auth/auth/scopes.py
```diff
@@ -0,0 +1,3 @@
+CUSTOMER_SCOPES = frozenset(["customer:free", "customer:paid"])
+PULSE3D_USER_SCOPES = frozenset(["pulse3d:free", "pulse3d:paid"])
+PULSE3D_SCOPES = [*CUSTOMER_SCOPES, *PULSE3D_USER_SCOPES]
```
tannermpeterson, line 3: This should also be immutable

core/lib/jobs/jobs/jobs.py
```diff
@@ -213,3 +217,30 @@ def _get_placeholders_str(num_placeholders, start=1):
...
+    """
+
+    async with con.transaction():
+        # get service specific usage restrictions for the customer account
+        usage_query = "SELECT usage_restrictions->$1 AS usage FROM customers WHERE id=$2"
+        usage_json = await con.fetchrow(usage_query, service, customer_id)
+        usage_dict = json.loads(usage_json["usage"])
+
+        # grab total uploads and jobs of customer for specific service
+        query = "SELECT COUNT(*) as total_uploads, SUM(jobs_count) as total_jobs FROM (SELECT COUNT(*) AS jobs_count FROM jobs_result WHERE customer_id=$1 AND type=$2 GROUP BY upload_id) dt"
```
tannermpeterson, line 237: 👍

core/lib/jobs/jobs/jobs.py
```diff
@@ -213,3 +217,30 @@ def _get_placeholders_str(num_placeholders, start=1):
...
+        customer_data = await con.fetchrow(query, customer_id, service)
+        total_uploads = customer_data["total_uploads"]
+        total_jobs = customer_data["total_jobs"] if customer_data["total_jobs"] is not None else 0
```
tannermpeterson, line 240: Would the `total_jobs` value returned by the query ever be `None`? I'm honestly not sure

deployments/apiv2/services/users/src/main.py
```diff
@@ -126,9 +129,25 @@ async def login(request: Request, details: Union[UserLogin, CustomerLogin]):
...
+                if is_customer_login_attempt:
+                    # get tier of service scope in list of customer scopes
+                    scope = [s for s in scope if details.service in s]
+                    # replace with customer scope
+                    # raises IndexError if not found
+                    scope[0] = scope[0].replace(details.service, "customer")
+
+                tokens = await _create_new_tokens(con, row["id"], customer_id, scope, account_type)
+                return LoginResponse(tokens=tokens, usage_quota=usage_quota)
+    except IndexError:
```
tannermpeterson, line 147: I think it would be better to catch the index error right where it happens and then raise a login error with this message instead, then it can just get handled by `except LoginError as e:`

tannermpeterson: Or you could just do
```if not scope:
   raise LoginError(msg)
```

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -250,11 +272,15 @@ async def _get_jobs(con, token, job_ids):
...
-    request: Request, details: JobRequest, token=Depends(ProtectedAny(scope=["users:free"]))
+    request: Request,
+    details: JobRequest,
+    token=Depends(ProtectedAny(scope=PULSE3D_USER_SCOPES)),
 ):
     try:
         user_id = str(uuid.UUID(token["userid"]))
-        logger.info(f"Creating pulse3d job for upload {details.upload_id} with user ID: {user_id}")
+        customer_id = str(uuid.UUID(token["customer_id"]))
+        service = token["scope"][0].split(":")[0]
```
tannermpeterson, line 282: It seems like with the new scopes we do a lot of splitting on ":", replacing etc. It would probably be good to make some utility functions that do this that we put in one of the core libs so we don't accidentally introduce a bug just in one place by splitting incorrectly or something else

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -291,18 +317,27 @@ async def create_new_job(
 
         priority = 10
         async with request.state.pgpool.acquire() as con:
+
+            usage_quota = await check_customer_quota(con, customer_id, service)
+            if usage_quota["jobs_reached"]:
+                raise CustomerUsageQuotaReached()
```
tannermpeterson, line 323: If this error is just going to get caught and then immediately return the response, we might as well just return the `UsageErrorResponse(usage_error=usage_quota)` here. I do think it would be useful though to include some logging so we can see in the svc logs if people are hitting routes and getting the `UsageErrorResponse`, so with that addition I think you can do it either way

core/lib/jobs/jobs/jobs.py
luciipak: New users get None returned from the SUM if there is nothing to sum

deployments/pulse3d/services/pulse3d/src/main.py
luciipak: Ah good catch! I had initially done that when I was returning an error code instead of changing the response body with 200

tannermpeterson (approved):

PR #287, opened 2022-10-29
permissions fix

luciipak:


core/db/curibio/alembic/versions/da5a93c1bb08_update_uploadtype_and_scope_schema.py
```diff
@@ -43,6 +43,11 @@ def upgrade():
             server_default="pulse3d",
         ),
     )
+    op.execute("GRANT ALL PRIVILEGES ON TABLE jobs_result TO curibio_users")
```
jperson, line 46: why doesn't `curibio_users` need all privileges on `jobs_result`? Same for `curibio_jobs` and `customers table?

luciipak: @jperson fixed to just select privileges

core/db/curibio/alembic/versions/da5a93c1bb08_update_uploadtype_and_scope_schema.py
```diff
@@ -58,9 +63,12 @@ def downgrade():
...
     op.execute("""ALTER TABLE uploads ALTER COLUMN type TYPE "UploadType" USING type::text::"UploadType" """)
-    op.execute(
-        """ALTER TABLE jobs_result ALTER COLUMN type TYPE "UploadType" USING type::text::"UploadType" """
-    )
+
     # drop the old type, cleanup
     op.execute("DROP TYPE oldUploadType")
     op.execute("UPDATE uploads SET type='mantarray' where type='nautilus'")
+
+    op.execute("REVOKE ALL PRIVILEGES ON TABLE jobs_result FROM curibio_users")
```
jperson, line 71: these should probably only revoke the `select` privileges

luciipak: ah, thanks

frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -433,7 +433,7 @@ export default function UploadForm() {
     <Container>
       <Uploads>
         <Header>Run Analysis</Header>
-        {tabSelection === "0" ? (
+        {tabSelection === "Analyze New Files" ? (
```
tannermpeterson, line 436: Should this be the same as the check that got changed above? They were different before

luciipak: the difference in indices is what's causing the issue you're seeing. Only one of them was changed in the PR that swaps the options

tannermpeterson (approved):

PR #288, opened 2022-10-31
added prefix to parquet file

NikitaPoly:
Added pulse3dVersion to parquet file name


tannermpeterson (changes requested):
I think we want the prefix to be separated by a `/` not a `_`

NikitaPoly:
I will make this change, why is `/` better in this situation ? @tannermpeterson

tannermpeterson:
It will allow S3 will recognize it as a prefix

tannermpeterson (approved):

PR #293, opened 2022-10-31
Json fe bump

NikitaPoly:
bumped version of FE to match s3


luciipak:
the version wasn't changed

NikitaPoly:
@luciipak  I added the change to the json, do I need to make it in a different file?

luciipak:
the diff only shows the package-lock.json was changed? should be in the package.json

tannermpeterson:
You change the version package.json, then run you can run `npm i` to updated package-lock.json. You shouldn't ever edit package-lock.json manually

luciipak (approved):

PR #295, opened 2022-11-01
S3 parquet

NikitaPoly:
Added prefix checking when getting parquet file from s3

### DIFF
--- a/deployments/pulse3d/services/pulse3d/src/main.py
+++ b/deployments/pulse3d/services/pulse3d/src/main.py
@@ -478,6 +478,7 @@ async def get_interactive_waveform_data(
             logger.info(f"Downloading recording data from {key}")
 
             download_directory_from_s3(bucket=PULSE3D_UPLOADS_BUCKET, key=key, file_path=tmpdir)
+
             # read the time force dataframe from the parquet file
             # older versions of pulse3d to not inlcude the version data
             if hasattr(parsed_meta, "version"):
--- a/deployments/pulse3d/services/pulse3d/src/main.py
+++ b/deployments/pulse3d/services/pulse3d/src/main.py
@@ -4,10 +4,8 @@ import logging
 from typing import List, Optional, Tuple, Union
 import uuid
 import tempfile
-from attr import has
 import boto3
 import os
-from importlib_metadata import version
 import numpy as np
 import pandas as pd
 from glob import glob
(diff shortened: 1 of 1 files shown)


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -480,7 +480,14 @@ async def get_interactive_waveform_data(
             download_directory_from_s3(bucket=PULSE3D_UPLOADS_BUCKET, key=key, file_path=tmpdir)
 
             # read the time force dataframe from the parquet file
-            parquet_path = glob(os.path.join(tmpdir, "time_force_data", "*.parquet"), recursive=True)
+            # older versions of pulse3d to not inlcude the version data
+            if hasattr(parsed_meta, "version"):
```
tannermpeterson, line 484: I think this should be a `dict` in which you can check if a key is present by just doing `if "version" in parsed_meta`. It would be best though in this case to just use `parsed.get("version")` combined with the walrus operator `if pulse3d_version := parsed_meta.get("version"):`.

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -480,7 +480,14 @@ async def get_interactive_waveform_data(
             download_directory_from_s3(bucket=PULSE3D_UPLOADS_BUCKET, key=key, file_path=tmpdir)
 
             # read the time force dataframe from the parquet file
-            parquet_path = glob(os.path.join(tmpdir, "time_force_data", "*.parquet"), recursive=True)
+            # older versions of pulse3d to not inlcude the version data
```
tannermpeterson, line 483: typo `include`

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -480,7 +480,14 @@ async def get_interactive_waveform_data(
...
+            if hasattr(parsed_meta, "version"):
+                parquet_path = glob(
+                    os.path.join(tmpdir, "time_force_data/", parsed_meta["version"], "*.parquet"),
```
tannermpeterson, line 486: The point of using `os.path.join` is so that you don't need to include a `/` or `\`, this function will determine the correct slash to use for you, so we can remove it here

tannermpeterson (changes requested):

tannermpeterson (approved):

PR #296, opened 2022-11-01
scroll no longer needed for table

NikitaPoly:
-uploads Table
  -- checkbox for uploads will also be disabled if jobs are not complete
  -- filter and sort are now in one section
  -- reverted styles to look more unified
  --actions drop down is now part of table
-Users Table
 -- Uses checkbox for selection
 --  now has same style as uploads table
 -- change coloring of columns based on `suspended `
 -- added resizable columns 
-Over all
 -- Fixed dashboard deselecting active menu item on refresh
 --  Can now press enter on log in screen


tannermpeterson (changes requested):
Need to add this change to every page

tannermpeterson (approved):

PR #301, opened 2022-11-04
added error modal

NikitaPoly:
Uploading a file with `(` or `)` will show an error to the user and prevent uploading.

### DIFF
--- a/frontend/pulse3d/src/pages/upload-form.js
+++ b/frontend/pulse3d/src/pages/upload-form.js
@@ -136,6 +136,19 @@ export default function UploadForm() {
   };
 
   useEffect(() => {
+    if (files.length > 0) {
+      const incompatible_names = [];
+      files.forEach((file) =>
+        file.name.includes("(") || file.name.includes(")") ? incompatible_names.push(file.name) : null
+      );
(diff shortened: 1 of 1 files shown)


frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -135,6 +135,19 @@ export default function UploadForm() {
     setCheckedParams(newCheckedParams);
   };
 
+  useEffect(() => {
+    if (files.length > 0) {
+      const incompatible_names = [];
+      files.forEach((file) =>
+        file.name.includes("(") || file.name.includes(")") ? incompatible_names.push(file.name) : null
+      );
+      if (incompatible_names.length > 0) {
```
luciipak, line 144: You could filter the files for incompatible names to simplify:
`const incompatible_names = files.filter((file) => file.name.includes("(") || file.name.includes(")"));`

NikitaPoly: So instead of prompting the user to reupload we just remove the conflicting files?

jperson: we want to allow users to upload files with those characters since they are valid file names

PR #304, opened 2022-11-08
Fe qc changes

NikitaPoly:


frontend/pulse3d/src/components/table/ColumnHead.js
```diff
@@ -23,41 +23,82 @@ export default function ColumnHead({
...
   }, [input]);
 
   const updateWidth = (e) => {
-    const difference = e.clientX - initialX;
+    const difference = parseInt((e.clientX - initialX) / 10);
     const newWidth = parseInt(width) + difference;
     const newNeighborWidth = parseInt(rightWidth) - difference;
     //prevent user from making columns to small
-    if (newWidth < 150 || newNeighborWidth < 150) {
+    // if (newWidth < 150 || newNeighborWidth < 150) {
```
luciipak, line 52: Remove commented code if no longer needed

frontend/pulse3d/src/components/table/ColumnHead.js
```diff
@@ -23,41 +23,82 @@ export default function ColumnHead({
...
+      {order === "" ? (
+        <EjectIcon
+          onClick={() => {
+            setOrder("asc");
+            setSortColumns(columnName);
+          }}
+          style={{ color: "gray" }}
+        />
+      ) : null}
+      {order === "asc" && (
```
luciipak, line 86: Does this return the `EjectIcon` with an && conditional? I'm more curious, I didn't know you could do that

frontend/pulse3d/src/components/table/UploadsSubTable.js
```diff
@@ -4,33 +4,52 @@ import Checkbox from "@mui/material/Checkbox";
...
 `;
-const SubRow = styled.div`
-  font-size: 0.75rem;
-  margin: 1rem 0;
+
+const FilenameHeader = styled.div`
+  padding-left: 3.9%;
+  width: 40%;
+`;
+const DateHeader = styled.div`
```
luciipak, line 28: Couldn't you reuse a header styled div  at 20% instead of having three of the same?

frontend/pulse3d/src/components/table/UploadsSubTable.js
```diff
@@ -4,33 +4,52 @@ import Checkbox from "@mui/material/Checkbox";
...
 `;
+
 const SubRowFileName = styled.div`
   font-size: 0.75rem;
-  margin: 1rem 0;
+  width: 40%;
 `;
-const FilenameHeader = styled.div`
-  width: 20rem;
+const SubRowDate = styled.div`
```
luciipak, line 42: same here with reusing

frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -422,7 +421,6 @@ export default function UploadForm() {
   const handleClose = async (idx) => {
     // if user chooses to proceed with upload when some files were flagged as bad
     if (idx === 1) {
-      const filteredFiles = files.filter((f) => !failedUploadsMsg.includes(f.name));
```
luciipak, line 425: was this meant to be removed?

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -130,24 +129,18 @@ export default function Uploads() {
...
-  const [uploadWidth, setUploadWidth] = useState("250px");
-  const [createdWidth, setCreatedWidth] = useState("250px");
-  const [analyzedWidth, setAnalyzedWidth] = useState("200px");
-
+  const [ownerWidth, setOwnerWidth] = useState("10%");
+  const [recordingWidth, setRecordingWidth] = useState("30%");
+  const [uploadWidth, setUploadWidth] = useState("23%");
+  const [createdWidth, setCreatedWidth] = useState("15%");
+  const [analyzedWidth, setAnalyzedWidth] = useState("15%");
+  const [sortColumn, setSortColumns] = useState("");
```
luciipak, line 137: `sortColumns` or `setSortColumn`

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -130,24 +129,18 @@ export default function Uploads() {
...
       admin: false,
       cell: (row) => (
-        <Checkbox
-          id={row.id}
-          disabled={row.jobs.filter((job) => job.status !== "finished").length > 0}
-          checked={checkedUploads.includes(row.id)}
-          onChange={handleCheckedUploads}
-          style={{ width: "1px", margin: "0" }}
-        />
+        <Checkbox id={row.id} checked={checkedUploads.includes(row.id)} onChange={handleCheckedUploads} />
```
luciipak, line 143: Did we decide to never disable the checkboxes or at least disable then while in pending state?

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -17,15 +18,35 @@ const customStyles = {
...
+  cells: {
+    style: { padding: "0 0 0 1.3%" },
+  },
 };
 
+const modalObjs = {
+  deactivate: {
+    header: "Attention",
+    messages: [
+      "You are trying to deactivate some already inactive accounts. They will be ignored. Error Users :",
```
luciipak, line 46: I might change this  to 'You are attempting to deactivate accounts that are already inactive. ' or something

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -303,13 +343,7 @@ export default function UserInfo() {
...
-                    checkedUsers.length === 0,
-                    checkedUsers.length === 0 ||
-                      usersData
-                        .filter((user) => checkedUsers.includes(user.id))
-                        .filter((checkedUsers) => checkedUsers.suspended === "suspended").length !== 0,
-                  ]}
+                  disableOptions={[checkedUsers.length === 0, checkedUsers.length === 0]}
                   optionsTooltipText={[
                     "Must make a selection below before actions become available.",
                     "Must select a user who is active before actions become available.",
```
luciipak, line 349: Do you need the modal warning about attempting to deactivate existing   inactive users if this tooltip says must select for active users? 
Since you're removing the conditional check above ensuring they aren't suspended, I would change this tooltip message or put the disable conditional back in and remove the warning modal.

frontend/pulse3d/src/pages/users-info.js
```diff
@@ -322,6 +356,15 @@ export default function UserInfo() {
           />
         </Container>
       </PageContainer>
+      <ModalWidget
+        open={modalState === "generic"}
+        labels={modalLabels.messages}
+        buttons={modalButtons}
+        closeModal={() => {
+          setModalState("");
```
luciipak, line 364: You might also need to reset the deactivate modal messages because you're pushing usernames into the array

frontend/pulse3d/src/components/table/ColumnHead.js
NikitaPoly: Yes. This is JSX specific I think.  https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

luciipak: Yeah that's cool, I'll start doing that

frontend/pulse3d/src/pages/users-info.js
NikitaPoly: I removed the modal and added the disable check back

frontend/pulse3d/src/pages/uploads.js
```diff
@@ -130,23 +129,22 @@ export default function Uploads() {
...
+      width: "3%",
       admin: false,
       cell: (row) => (
         <Checkbox
           id={row.id}
-          disabled={row.jobs.filter((job) => job.status !== "finished").length > 0}
           checked={checkedUploads.includes(row.id)}
           onChange={handleCheckedUploads}
-          style={{ width: "1px", margin: "0" }}
+          disabled={row.jobs.filter((job) => job.status == "pending").length > 0}
```
luciipak, line 147: Sorry I think I misspoke on this, the individual jobs are disabled as expected with `disabled={job.status === "pending"}`, so we don't want any restrictions here on the actual uploads.. so what you had was right. We just want to make sure that the modal pops up as expected if they select an upload and there is a pending/error jobs

NikitaPoly: I added this back in. The modal seems to be working but I will check in my errors pr

luciipak: We just want to make sure the corresponding modal pops up if:
-  they select all and the upload has a pending job if attempting to download or delete 
-  they select all and has an error job if attempting to download

NikitaPoly: I see. Just confirmed modal does pop up in both situations. A count of succeeded files is shown too.

luciipak: Actually, I don't think there is a modal for the first case.. so does something pop up?
I guess is fine since the checkbox is clearly disabled. Might want to prevent deleting of an entire upload tree if it contains a pending job because I'm not sure what will happen if they try. Probably just the job would error if it attempts to update the jobs_result table and the entry is gone

NikitaPoly: No matter what you select, it will only download completed jobs. If there are none then it will tell the user no valid files to download/  delete selected.

luciipak: Okay cool, sorry just wanted to double check. So if a user selects an upload and all jobs, but it contains a pending job, it won't let them delete correct?

NikitaPoly: It does let them delete. I think I see what your saying now. If there is a pending job, it should error and not take action?

NikitaPoly: or remove the pending jobs and still take action?

luciipak: We don't want them to be able to delete the upload that contains a pending job

NikitaPoly: Can I add it to this ticket? https://app.asana.com/0/1152302614804188/1203331640504342/f
This ticket is about  fixing some bugs with alert modals.

luciipak (approved):

PR #305, opened 2022-11-08
Case insensitive

NikitaPoly:
New usernames will all be lowercase while all old usernames set to be lowercase.


luciipak:
Does it work for users logging in? like if they log in with a different case from what's in the db

deployments/apiv2/services/users/src/main.py
```diff
@@ -408,7 +408,7 @@ async def get_all_users(
     customer_id = uuid.UUID(hex=token["userid"])
 
     query = (
-        "SELECT id, name, email, created_at, last_login, suspended FROM users "
+        "SELECT id, LOWER(name), email, created_at, last_login, suspended FROM users "
```
luciipak, line 411: Is the `LOWER` for extra safety measures? They should all be lowercase in the DB correct?

NikitaPoly: I guess it is redundant. I will remove it

deployments/apiv2/services/users/src/main.py
```diff
@@ -454,7 +454,7 @@ async def get_user(
...
```
luciipak, line 457: this might also be redundant then?

luciipak:
And does this work with all the tests?

NikitaPoly:
All tests are good. Ran them again just to make sure

luciipak (approved):

PR #308, opened 2022-11-09
Modal errors

NikitaPoly:


frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -290,20 +302,19 @@ export default function UploadForm() {
...
-      if (badZipfiles.length > 0) {
-        // give users the option to proceed with clean files if any, otherwise just close
-        setModalButtons(badZipfiles.length !== files.length ? ["Cancel", "Proceed"] : ["Close"]);
-
-        // add files to modal to notify user which files are bad
-        setFailedUploadsMsg([defaultZipErrorLabel, ...badZipfiles.map((f) => f.name)]);
-
-        setModalState(true);
-        return;
+      setBadZipFiles(badZipfiles);
```
luciipak, line 305: Is this getting reset anywhere? Want to make sure that if they try again, this is reset to empty

NikitaPoly: it resets it at the beginning of each iteration overwriting the last. So if there are no ne badzipfiles it will reset it with `[]`. if there are bad zip files it will reset it to those zip files.

NikitaPoly: So you would not need to explicitly call a reset

luciipak (approved):

PR #314, opened 2022-11-14
ms converted to s on server

NikitaPoly:


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -388,7 +388,7 @@ def _format_tuple_param(
         return None
 
     if isinstance(default_values, int):
-        default_values = (default_values,) * len(options)
+        default_values = (default_values / 1000,) * len(options)
```
tannermpeterson, line 391: I don't think we want to touch the default values here as they are coming directly from pulse3d. Also, I don't think any of the values passed into this function are in units of time. Also are we sure we still need to do this conversion?

NikitaPoly: This would convert all the default values. I would have to change that. Right now pulse3d takes the units for width factor as ms. On the spread sheet the units are seconds. I'm guessing this is converted in pulse3d. We would not need to make changes?

PR #315, opened 2022-11-15
form reload on reset

NikitaPoly:
The component used for dragging files keeps its own state of files. When resetting the form, everything gets reset except the the internal state of the file drag component (this component is imported from a library). So when uploading a new file of a different name all functions as should. Uploading a file of the same name does not work because the internal state of the drag component does not change, does not fire the file change event.  Looking at the docs of the library, it does not seem that they have that level of control for external users.  Instead I reload the page when the user clicks reset.


jperson:
will reloading the page wipe out the auth token in the service worker?

NikitaPoly:
@jperson Yes I checked the service worker and all is working fine

PR #317, opened 2022-11-15
Reset drag drop

luciipak:


frontend/pulse3d/src/components/uploadForm/FileDragDrop.js
```diff
@@ -46,22 +47,30 @@ export default function FileDragDrop({
...
-        types={fileTypes}
-        multiple={true}
-      >
-        <DropZone style={dropZoneStyle}>
-          {dropZoneText}
-          <br />
-          <FileSelectionLabel>[ {fileSelection || "No file selected"} ]</FileSelectionLabel>
-        </DropZone>
-      </FileUploader>
+      {!resetDragDrop && (
```
NikitaPoly, line 59: To reset the state of the component it is enough to just re-render it?

luciipak: The conditional is essentially removing and then re-appending to the DOM again so it's force resetting the internal state of the file-drag-drop-files component

NikitaPoly (approved):

PR #320, opened 2022-11-18
Adv user scope

luciipak:


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -532,10 +592,15 @@ async def get_interactive_waveform_data(
                     os.path.join(tmpdir, "time_force_data", pulse3d_version, "*.parquet"),
                     recursive=True,
                 )
+                df = pd.read_parquet(parquet_path)
+                # some files have version but parquet data is not stored under it
+                if df.empty:
```
luciipak, line 597: is there a better way to do this? 
some entries in the database have a version in the metadata but the original parquet is still stored at root directory?

jperson: Could this update the prefix of the parquet file so it matches the prefix schema for current jobs?

jperson (approved):

deployments/pulse3d/services/pulse3d/src/main.py
luciipak: What do you mean?

PR #329, opened 2022-12-02
Store peaks/valleys in S3

luciipak:


deployments/pulse3d/services/pulse3d/src/requirements.txt
```diff
@@ -10,7 +10,6 @@ starlette==0.16.0
 typing-extensions==4.0.1
 uvicorn[standard]==0.16.0
 stream-zip==0.0.48
-pulse3d==0.27.0
```
NikitaPoly, line 13: Should this be updated to a new version instead of deletion?

luciipak: @NikitaPoly The reason we want to store peaks and valleys in parquet files is to remove this dependency from the pulse3d service so that we don't have to worry about the various versions. So this PR removes anything being used from Pulse3D and is no longer being imported for anything. It was initially only being used for `peak_detector` in interactive analysis

tannermpeterson: I think it makes sense to keep it as an import so we can keep a single source of truth for the constants that we import from it. But then again, these could change in new pulse3d versions, so it might actually make more sense to not have them in this svc at all and grab them in the job instead

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -423,6 +424,26 @@ async def create_new_job(
                 job_type=service,
             )
 
+            # NOTE this happens after the job is already created to have access to the job id, hopefully this doesn't cause any issues with the job starting before the file is uploaded to s3
```
tannermpeterson, line 427: I think you should probably do `# Luci (<date>): this happens ...` to keep it consistent with our other repos

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -423,6 +424,26 @@ async def create_new_job(
...
+            # Versions less than 0.28.0 should not be in the dropdown as an option, this is just an extra check for versions greater than 0.28.0
```
tannermpeterson, line 428: Is this saying that we have to deprecate all versions before 0.28.0 for new analyses or just interactive analysis? People might still be using older versions for both so I'm not sure that we can do that for either right now

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -559,17 +580,22 @@ async def get_interactive_waveform_data(
...
         selected_job = jobs[0]
         parsed_meta = json.loads(selected_job["job_meta"])
-        analysis_params = parsed_meta["analysis_params"]
         recording_owner_id = str(selected_job["user_id"])
+        pulse3d_version = parsed_meta.get("version", None)
+
+        # only compatible with Pulse3D 0.28.0 and above because of changes in pulse3d-worker
+        # interactive analysis was initially compatible with 0.25.2 and above, but pulse3d-worker is not backwards compatible
+        if pulse3d_version is None or VersionInfo.parse(pulse3d_version) < "0.28.0":
+            return NoJobDataFoundResponse(
```
tannermpeterson, line 596: We probably want this to return some kind of 4xx error code, I don't think it does that right now. Also once we do that I'm not sure if we need to have an entire Model just to return an error message

deployments/pulse3d/services/pulse3d/tests/test_main.py
```diff
@@ -769,14 +831,14 @@ def test_jobs__post__omits_analysis_params_not_supported_by_the_selected_pulse3d
...
         expected_analysis_param_keys.append("max_y")
-    if pulse3d_semver >= "0.25.2":
-        expected_analysis_param_keys.append("peaks_valleys")
     if pulse3d_semver >= "0.25.4":
         expected_analysis_param_keys.append("normalize_y_axis")
     if pulse3d_semver >= "0.26.0":
         expected_analysis_param_keys.append("stiffness_factor")
     if pulse3d_semver >= "0.27.4":
         expected_analysis_param_keys.append("inverted_post_magnet_wells")
+    if "0.25.2" <= pulse3d_semver < "0.28.0":
```
NikitaPoly, line 840: Peaks and valleys will not be supported in versions later than 0.28.0?

luciipak: On line 144 in `/pulse3d/src/main.py`, this is where peaks and valleys get handled with versions greater than 0.28.0. This line `"0.25.2" <= pulse3d_semver < "0.28.0"` allows compatibility with older pulse3d versions so that users can still use it with recordings analyzed with older versions.

frontend/pulse3d/src/serviceWorker.js
```diff
@@ -222,7 +222,8 @@ const interceptResponse = async (req, url) => {
       const resBodyToCheck = await response.json();
 
       // set the usage error to SW state to send in auth check, will return a 200 status
-      if (resBodyToCheck.usage_error) setUsageQuota(resBodyToCheck.usage_error);
+      if (resBodyToCheck.error && resBodyToCheck.error === "UsageError")
```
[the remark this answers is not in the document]
luciipak: and I want to return `new Response(JSON.stringify(resBodyToCheck))` in all cases

NikitaPoly (approved):

PR #330, opened 2022-12-06
first change

NikitaPoly:
kustomize changes

### DIFF
--- a/deployments/pulse3d/manifests/base/pulse3d-dep.yaml
+++ b/deployments/pulse3d/manifests/base/pulse3d-dep.yaml
@@ -19,7 +19,6 @@ spec:
     spec:
       containers:
       - name: pulse3d
-        image: 077346344852.dkr.ecr.us-east-2.amazonaws.com/pulse3d_api:0.5.0
         env:
         - name: POSTGRES_USER
           value: curibio_jobs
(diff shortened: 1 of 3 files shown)


deployments/pulse3d/manifests/base/pulse3d-dep.yaml
```diff
@@ -19,7 +19,6 @@ spec:
     spec:
       containers:
       - name: pulse3d
-        image: 077346344852.dkr.ecr.us-east-2.amazonaws.com/pulse3d_api:0.5.0
```
luciipak, line 22: we probably need to add this to the test overlay kustomization since there is no default now

deployments/pulse3d/manifests/overlays/test/kustomization.yaml
```diff
@@ -6,3 +6,27 @@ resources:
...
+    - op: replace
+      path: /spec/rules/0/host
+      value: "pulse3d.curibio-test.com"
+- target:
+    kind: Deployment
+    name: pulse3d
+  patch: |-
+    - op: add
+      path: /spec/template/spec/containers/0/image
+      value: "077346344852.dkr.ecr.us-east-2.amazonaws.com/pulse3d_api:0.5.0"
```
luciipak, line 24: You really only need to add this one since it's the one you removed from pulse3d-dep.yaml. The other values default to the test cluster and that's why they get replaced/added in the prod overlay

luciipak (approved):

PR #331, opened 2022-12-06
restricted jobs to run for 15 mins max

NikitaPoly:


jperson:
this changes the runtime for workflows, we want to restrict the pulse3d jobs to 15 minutes

jperson:
this needs to be the pulse3d-worker, not the pulse3d api

PR #332, opened 2022-12-07
Added reset password for customer accounts

luciipak:


deployments/apiv2/services/users/src/main.py
```diff
@@ -405,14 +413,23 @@ async def _create_user_email(
...
     scope: List[str],
-    name: str,
+    name: Optional[str],
     email: EmailStr,
 ):
     try:
+        if "user" in scope[0]:
+            account_type = "user"
+            query = "UPDATE users SET pw_reset_verify_link=$1 WHERE id=$2"
+        else:
```
jperson, line 425: this is nitpicky but can you make this so it explicitly checks for each type so that if there's a bug that somehow creates a scope with an `account_type` other than `users` it doesn't allow customer pw resets. Something like

```
if "user" in scope[0]:
    ...
elif "customer" in scope[0]:
    ...
else:
    return an error here
```

This is mostly for being defensive in the code, I don't think it's likely that could happen but if there was a typo in a PR that created say a scope for `users` this would allow the customer password to be changed instead.

luciipak: @jperson yep got it!

deployments/apiv2/services/users/src/main.py
```diff
@@ -484,33 +499,52 @@ async def verify_user_account(
...
+                    if row["pw_reset_link"] is None:
+                        return UnableToUpdateAccountResponse(message="Link has already been used")
+
+                    await con.execute(
+                        "UPDATE customers SET pw_reset_link=NULL, password=$1 WHERE id=$2",
+                        phash,
+                        user_id,
+                    )
+
+                elif is_user:
```
tannermpeterson, line 527: Is there ever a situation where the account is not a user or customer? If it's always one of these two, you could just make this `else:` since you know it's a user if it's not a customer

luciipak: @tannermpeterson  in the resolved convo, jason mentioned wanting to prevent other types and to write it more defensively so only accept customer and users, don't use a catchall else statement

PR #337, opened 2022-12-08
user table updates on login

NikitaPoly:


deployments/apiv2/services/users/src/main.py
```diff
@@ -154,6 +160,9 @@ async def login(request: Request, details: Union[UserLogin, CustomerLogin]):
                     _, customer_tier = split_scope_account_data(scope[0])
                     scope[0] = f"customer:{customer_tier}"
 
+                # after login flow succes update the last-login filed of the user
```
tannermpeterson, line 163: typo: success. Also not sure if "filed" is supposed to be "field", but this whole comment should probably read: if login was successful, that update last_login column to now

deployments/apiv2/services/users/src/main.py
```diff
@@ -103,6 +103,9 @@ async def login(request: Request, details: Union[UserLogin, CustomerLogin]):
         )
         select_query_params = (details.email,)
         customer_id = None
+
+        update_last_login_query = "UPDATE users SET last_login = $1 WHERE deleted_at IS NULL AND email = $2"
```
luciipak, line 107: you'll want to update the customers table here

luciipak (approved):
Make sure to push a new image with changes

PR #342, opened 2022-12-13
added modal

NikitaPoly:
Added a a check before interactive analysis can be submitted for duplicate peaks or valleys

### DIFF
--- a/frontend/pulse3d/src/components/uploads/InteractiveAnalysisModal.js
+++ b/frontend/pulse3d/src/components/uploads/InteractiveAnalysisModal.js
@@ -157,6 +157,7 @@ export default function InteractiveWaveformModal({ selectedJob, setOpenInteracti
   const [changelog, setChangelog] = useState({});
   const [openChangelog, setOpenChangelog] = useState(false);
   const [undoing, setUndoing] = useState(false);
+  const [duplicatesPresent, setDuplicatesPresent] = useState(false);
 
   const [xRange, setXRange] = useState({
     min: null,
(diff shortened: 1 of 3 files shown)


luciipak:
This is great! quick thing,  I noticed if you add more than one contraction or relaxation point, it only marks the first one, it does correct it once you delete the red one. and it also doesn’t stay red when you move it

frontend/pulse3d/src/components/uploads/InteractiveAnalysisModal.js
```diff
@@ -117,6 +117,11 @@ const constantModalLabels = {
     messages: ["There was an issue while attempting to start this analysis.", "Please try again later."],
     buttons: ["Close"],
   },
+  duplicate: {
+    header: "Before Running Analysis!",
+    messages: ["Check no double peaks or valleys exist", "(Red Triangles)"],
```
luciipak, line 122: This might be better to say something like 'They are shown as red triangles above.' and  add a period at the end of the first line

frontend/pulse3d/src/components/uploads/WaveformGraph.js
```diff
@@ -444,8 +495,12 @@ export default function WaveformGraph({
       .attr("transform", (d) => {
         return "translate(" + x(dataToGraph[d][0]) + "," + (y(dataToGraph[d][1]) - 7) + ") rotate(180)";
       })
-      .style("fill", "orange")
-      .attr("stroke", "orange")
+      .style("fill", (d) => {
```
luciipak, line 498: You can change the color of the triangles while dragging in the function on line 429 so that they stay red

luciipak (approved):

PR #347, opened 2022-12-15
Link to metric definitions

NikitaPoly:
Added a button under `Account Settings` that opens a new tab with the pulse3d docs.

https://user-images.githubusercontent.com/46114925/207988306-e96c7677-6ec7-4ace-b936-454689aeb931.mp4

### DIFF
--- a/frontend/pulse3d/src/components/layouts/ControlPanel.js
+++ b/frontend/pulse3d/src/components/layouts/ControlPanel.js
@@ -145,6 +145,7 @@ export default function ControlPanel() {
       page: "/account",
       options: [],
     },
+    { label: "Metric Definitions", options: [] },
   ];
 
   const buttons = accountType === "admin" ? adminButtons : userButtons;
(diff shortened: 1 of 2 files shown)


frontend/pulse3d/src/components/layouts/ControlPanel.js
```diff
@@ -190,6 +191,11 @@ export default function ControlPanel() {
 
           const handleSelected = (e) => {
             e.preventDefault();
+            //if metric definitions button then open window with metric definitions
+            if (label === userButtons[3].label) {
+              window.open("https://pulse3d.readthedocs.io/en/latest/gettingstarted.html#jupyter-notebooks");
```
tannermpeterson, line 196: @jperson @luciipak is there anything in the CSP that would prevent this from opening? I'm assuming the answer is no right now since Nikita posted a video of it working, but just wondering if we'll need to add this to the CSP in the future

NikitaPoly: This is also on the test cluster

tannermpeterson (approved):

PR #349, opened 2022-12-17
duplicate features wont prevent analysis  

NikitaPoly:


frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -720,9 +720,8 @@ export default function InteractiveWaveformModal({ selectedJob, setOpenInteracti
                 if (duplicatesPresent) {
                   setModalLabels(constantModalLabels.duplicate);
                   setModalOpen("duplicate");
-                } else {
-                  postNewJob();
                 }
+                postNewJob();
```
luciipak, line 724: I think they should be given the option to close the modal and make changes before confirming the job submission. I wouldn't submit the job until the modal has been handled

luciipak: If we submit the job here and as this modal appears, the job might become successful and another modal auto appears when a job is submitted so it would close this modal. It might be confusing. SO I would make this have 'Close' and 'Continue' buttons. If 'Continue', submit job, else if 'Close' then close modal and let the user keep editing

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -121,9 +121,9 @@ const constantModalLabels = {
     buttons: ["Close"],
   },
   duplicate: {
-    header: "Before Running Analysis!",
-    messages: ["Check no double peaks or valleys exist.", "They are shown as red triangles above."],
-    buttons: ["Close"],
+    header: "Warning!",
+    messages: ["Duplicate peaks and valleys detected."],
+    buttons: ["Continue", "Cancel"],
```
luciipak, line 126: nit-picky, but all the modals through the desktop app and dashboard have 'cancel' or 'close' on the left. Probably just want to keep them the same in case people do it out of habit

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -162,7 +162,18 @@ export default function InteractiveWaveformModal({ selectedJob, setOpenInteracti
...
   const [duplicatesPresent, setDuplicatesPresent] = useState(false);
-
+  const [duplicateModalOpen, setDuplicateModalOpen] = useState(false);
+
+  const handleDuplicatesModalClose = (choice) => {
+    //0 = cancel
+    //1 = continue
+    if (choice === 0) {
+      setDuplicateModalOpen(false);
+    } else if (choice === 1) {
```
tannermpeterson, line 172: The comment you left is a good indication that `choice` could just be renamed to one of the two possible options, `cancel` or `continue`. Also this Boolean logic should be refactored since `setDuplicateModalOpen(false);` is present in both (all) branches:
```
setDuplicateModalOpen(false);
if (continue) {
 postNewJob();
}
```

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -121,9 +121,9 @@ const constantModalLabels = {
     buttons: ["Close"],
   },
   duplicate: {
-    header: "Before Running Analysis!",
-    messages: ["Check no double peaks or valleys exist.", "They are shown as red triangles above."],
-    buttons: ["Close"],
+    header: "Warning!",
+    messages: ["Duplicate peaks and valleys detected."],
```
tannermpeterson, line 125: This message is a little ambiguous, someone may think it means that both duplicate peaks and duplicate valleys exist. Could just change it to `"Duplicate peaks and/or valleys detected."`

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -162,7 +162,12 @@ export default function InteractiveWaveformModal({ selectedJob, setOpenInteracti
   const [undoing, setUndoing] = useState(false);
   const [peakValleyWindows, setPeakValleyWindows] = useState({});
   const [duplicatesPresent, setDuplicatesPresent] = useState(false);
+  const [duplicateModalOpen, setDuplicateModalOpen] = useState(false);
 
+  const handleDuplicatesModalClose = (choice) => {
+    setDuplicateModalOpen(false);
+    choice === 1 ? postNewJob() : null;
```
tannermpeterson, line 169: I would argue that using the ternary operator like this to conditionally run code is bad style. The most common use case you see for it is to conditionally assign a value to a variable, so when that is not the case the intent becomes unclear and makes the code less readable. Also, in this instance there's nothing to do if `choice` is falsey, but you are still forced to put a `null` in the statement which looks clunky.

Using a regular old `if` statement makes the intent immediately clear, and if you really want to keep it as a one liner you could do
```
if (choice) postNewJob();
```
Although I would also argue that
```
if (choice) {
  postNewJob();
}
```
is the most readable option. Also still think that `choice` should be renamed here to something that makes it obvious what action the user chose, like `runAnalysis`

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -121,9 +121,9 @@ const constantModalLabels = {
     buttons: ["Close"],
   },
   duplicate: {
-    header: "Before Running Analysis!",
-    messages: ["Check no double peaks or valleys exist.", "They are shown as red triangles above."],
-    buttons: ["Close"],
+    header: "Warning!",
+    messages: ["Consecutive peaks and/or valleys detected."],
+    buttons: ["Back to Edit", "Run Analysis"],
```
tannermpeterson, line 126: Can we make this `"Back to Editing"`?

tannermpeterson (changes requested):

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -725,12 +731,19 @@ export default function InteractiveWaveformModal({ selectedJob, setOpenInteracti
...
+        open={["status", "dataFound", "pulse3dWarning"].includes(modalOpen)}
         buttons={modalLabels.buttons}
         closeModal={handleModalClose}
         header={modalLabels.header}
         labels={modalLabels.messages}
       />
+      <ModalWidget
+        open={duplicateModalOpen}
+        buttons={constantModalLabels.duplicate.buttons}
+        closeModal={(userWarningChoice) => handleDuplicatesModalClose(userWarningChoice === 1)}
```
tannermpeterson, line 743: You don't need to do this, `1` is a truthy value and `0` is falsey so they work fine as conditions in `if` statements

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -725,12 +731,19 @@ export default function InteractiveWaveformModal({ selectedJob, setOpenInteracti
...
+        open={["status", "dataFound", "pulse3dWarning"].includes(modalOpen)}
         buttons={modalLabels.buttons}
         closeModal={handleModalClose}
         header={modalLabels.header}
         labels={modalLabels.messages}
       />
+      <ModalWidget
+        open={duplicateModalOpen}
+        buttons={constantModalLabels.duplicate.buttons}
+        closeModal={(userWarningChoice) => handleDuplicatesModalClose(userWarningChoice)}
```
tannermpeterson, line 743: `closeModal={handleDuplicatesModalClose}`

tannermpeterson (approved):

PR #350, opened 2022-12-21
Usage on register

luciipak:


core/lib/auth/auth/scopes.py
```diff
@@ -2,3 +2,4 @@
 PULSE3D_USER_SCOPES = frozenset(["pulse3d:free", "pulse3d:paid", "pulse3d:rw_all_data"])
 PULSE3D_SCOPES = frozenset([*CUSTOMER_SCOPES, *PULSE3D_USER_SCOPES])
 ACCOUNT_SCOPES = frozenset(["users:verify", "users:reset", "customer:reset"])
+PULSE3D_PAID_USAGE = {"pulse3d": {"uploads": -1, "jobs": -1}}
```
tannermpeterson, line 5: Can we make this immutable? You could just json.dumps it to a string instead of using immutabledict if that's nicer

luciipak: json.dumps here? it also json dumps in the route?

deployments/apiv2/services/users/src/main.py
```diff
@@ -305,11 +306,12 @@ async def register(
             phash = ph.hash(details.password1.get_secret_value())
             scope = details.scope
             # usage_restrictions column is not currently being inserted into, will need to be manually added
-            insert_query = "INSERT INTO customers (email, password, data) VALUES ($1, $2, $3) RETURNING id"
+            insert_query = "INSERT INTO customers (email, password, data, usage_restrictions) VALUES ($1, $2, $3, $4) RETURNING id"
             query_params = (
                 details.email,
                 phash,
                 json.dumps({"scope": scope}),
+                json.dumps(PULSE3D_PAID_USAGE),
```
tannermpeterson, line 314: It doesn't look like `json.dumps` likes `immutabledicts`, but `json.dumps(dict(PULSE3D_PAID_USAGE))` should work

luciipak: @tannermpeterson thanks for catching these, my bad for missing them. Christmas brain.

deployments/apiv2/services/users/tests/test_main.py
```diff
@@ -304,10 +304,11 @@ def test_register__customer__success(mocked_asyncpg_con, spied_pw_hasher, cb_cus
     }
 
     mocked_asyncpg_con.fetchval.assert_called_once_with(
-        "INSERT INTO customers (email, password, data) VALUES ($1, $2, $3) RETURNING id",
+        "INSERT INTO customers (email, password, data, usage_restrictions) VALUES ($1, $2, $3, $4) RETURNING id",
         registration_details["email"],
         spied_pw_hasher.spy_return,
         json.dumps({"scope": expected_scope}),
+        json.dumps(PULSE3D_PAID_USAGE),
```
tannermpeterson, line 311: This test will need to be updated too

tannermpeterson (approved):

PR #363, opened 2023-01-19
Add upload limits

NikitaPoly:


core/db/curibio/alembic/versions/2ccc9fe21ec1_adding_plan_end_date_to_usage_.py
```diff
@@ -0,0 +1,30 @@
...
+Revision ID: 2ccc9fe21ec1
+Revises: 1f23e047d2b0
+Create Date: 2023-01-25 14:40:08.612570
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "2ccc9fe21ec1"
```
luciipak, line 12: the database is at revision `89e819e0ecc2`, I would upgrade the database through alembic to make sure this works as expected. I think the dict needs another set of quotes to be in json

core/lib/jobs/jobs/jobs.py
```diff
@@ -219,28 +219,66 @@ def _get_placeholders_str(num_placeholders, start=1):
...
+    Returns:
+        - Dictionary with account limits and account usage
+    """
+    # get service specific usage restrictions for the customer account
+    # uploads limit, jobs limit, end date of plan
+    usage_limit_query = "SELECT usage_restrictions->$1 AS usage FROM customers WHERE id=$2"
+    # get total uploads of customer for specific service
+    current_usage_query = "SELECT COUNT(*) as total_uploads, SUM(jobs_count) as total_jobs FROM (SELECT COUNT(*) AS jobs_count FROM jobs_result WHERE customer_id=$1 AND type=$2 GROUP BY upload_id) dt"
+    # get all uploads with up to 2 jobs
+    less_than_3_jobs_query = "SELECT upload_id, COUNT(upload_id) FROM jobs_result WHERE customer_id=$1 GROUP BY upload_id HAVING COUNT(upload_id)<=2"
```
luciipak, line 233: We might want to also make sure that we only select from jobs for a specific product type like in the query on 231.. I haven't checked with product but I'm assuming there will be limits per product type.

core/db/curibio/alembic/versions/2ccc9fe21ec1_adding_plan_end_date_to_usage_.py
tannermpeterson: Yeah, we want to run migrations we add before posting a PR. There may be some exceptions of course, but something small like this we want to make sure works. It also helps other people working in the repo if you run your migration script right away so that when they make a new one using `alembic revision -m "<msg>"` it uses the correct `down_revision`

deployments/apiv2/services/users/src/main.py
```diff
@@ -3,6 +3,7 @@
 from typing import Union, List, Optional
 import uuid
 from datetime import datetime
+from jobs import check_customer_quota
```
luciipak, line 6: Can you run pre-commit checks? This will move back down to local imports

core/db/curibio/alembic/versions/2ccc9fe21ec1_adding_plan_end_date_to_usage_.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision = "1f23e047d2b0"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    new_usage_restrictions_struct = {"pulse3d": {"jobs": -1, "uploads": -1, "end": None}}
```
tannermpeterson, line 19: This should just be written directly into query instead of created as a python dict and inserted into an f-string

tannermpeterson: Also we should use a more descriptive name than "end", something like "expiration_date" would be better

deployments/pulse3d/services/pulse3d/src/models/models.py
```diff
@@ -11,9 +11,17 @@ class UploadRequest(BaseModel):
     upload_type: str
 
 
+class Usage_Quota(BaseModel):
```
tannermpeterson, line 14: 👍 +1

frontend/pulse3d/src/components/basicWidgets/UsageProgressWidget.js
```diff
@@ -0,0 +1,86 @@
+import { useEffect, useState } from "react";
+import CircularProgressWithLabel from "./CircularProgressWithLabel";
+import { useContext } from "react";
+import { AuthContext } from "@/pages/_app";
+export default function UsageProgressWidget({ labeltextcolor }) {
+  const { usageQuota } = useContext(AuthContext);
+  const [maxUploads, setMaxUploads] = useState(-1);
+  const [actualUploads, setActualUploads] = useState(0);
+  const [UsagePercentage, setUsagePercentage] = useState(0);
```
luciipak, line 9: usagePercentage

frontend/pulse3d/src/components/basicWidgets/UsageProgressWidget.js
```diff
@@ -0,0 +1,86 @@
...
+
+  useEffect(() => {
+    if (usageQuota) {
+      const limit = parseInt(usageQuota.limits.jobs);
+      const actual = parseInt(usageQuota.current.jobs);
+      setMaxUploads(limit);
+      setActualUploads(actual);
+      const UsagePercentage = parseInt((actual / limit) * 100);
```
luciipak, line 17: usagePercentage

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -157,8 +157,7 @@ async def create_recording_upload(
                     PULSE3D_UPLOADS_BUCKET,
                     upload_id=upload_id,
                 )
-
-                return UploadResponse(id=upload_id, params=params)
+                return UploadResponse(id=upload_id, params=params, usage_quota=usage_quota)
```
tannermpeterson, line 160: Why does usage need to be returned from this route?

frontend/pulse3d/src/components/basicWidgets/UsageWidgetFull.js
```diff
@@ -0,0 +1,51 @@
...
+            ? `${daysOfPlanLeft} days of plan left`
+            : `${daysOfPlanLeft * -1} days expired`}
+        </p>
+        <p>{`${actualUsage} out of ${limitUsage} ${metricName} used`}</p>
+        <CircularProgressWithLabel
+          value={(actualUsage / limitUsage) * 100 > 100 ? 100 : parseInt((actualUsage / limitUsage) * 100)}
+          labeltextcolor={labeltextcolor}
+        />
+        <p id="smallDescription">
+          Each upload comes with one free initial analysis and one free re-analysis. All subsequent analyses
```
luciipak, line 27: I thought the first analysis was counted and the first re-analysis wasn't? I ofc wasn't at the meeting for this, did we change it to a free analysis and re-analysis?

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -161,8 +165,10 @@ export default function InteractiveWaveformModal({ selectedJob, setOpenInteracti
   const [openChangelog, setOpenChangelog] = useState(false);
   const [undoing, setUndoing] = useState(false);
   const [peakValleyWindows, setPeakValleyWindows] = useState({});
+  const [duplicatesPresent, setDuplicatesPresent] = useState(false);
```
luciipak, line 168: I don't think this needs to be here anymore, duplicates is handled in the InteractiveAnalysis component

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -829,6 +840,14 @@ export default function InteractiveWaveformModal({ selectedJob, setOpenInteracti
             : ["No changes found."]
         }
       />
+      <ModalWidget
+        open={creditUsageAlert}
+        labels={["This Re-analysis will use 1 credit."]}
```
luciipak, line 845: re-analysis

frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -457,13 +469,18 @@ export default function UploadForm() {
...
+            <FileDragDrop // TODO figure out how to notify user if they attempt to upload existing recording
+              handleFileChange={(files) => setFiles(Object.values(files))}
+              dropZoneText={dropZoneText}
+              fileSelection={
+                files.length > 0 ? files.map(({ name }) => name).join(", ") : "No files selected"
+              }
+              setResetDragDrop={setResetDragDrop}
+              resetDragDrop={resetDragDrop}
+            />
+            <UploadCreditUsageInfo> Each file succesfully uploaded will use 1 credit.</UploadCreditUsageInfo>
```
luciipak, line 482: successfully*

frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -525,6 +542,14 @@ export default function UploadForm() {
         }}
         header={usageModalLabels.header}
       />
+      <ModalWidget
+        open={creditUsageAlert}
+        labels={["This Re-analysis will use 1 credit."]}
```
luciipak, line 547: re-analysis

frontend/pulse3d/src/serviceWorker.js
```diff
@@ -222,8 +225,15 @@ const interceptResponse = async (req, url) => {
       const resBodyToCheck = await response.json();
 
       // set the usage error to SW state to send in auth check, will return a 200 status
-      if (resBodyToCheck.error && resBodyToCheck.error === "UsageError")
+      if (resBodyToCheck.error && resBodyToCheck.error === "UsageError") {
         setUsageQuota(resBodyToCheck.message);
+      } else {
+        // most resent upload does not get read in time
```
luciipak, line 231: recent

deployments/pulse3d/services/pulse3d/src/main.py
NikitaPoly: I was anticipating this for when we add usage restrictions for uploads as well.

deployments/apiv2/services/users/src/main.py
NikitaPoly: I have been running pre commit but it did not catch this, I will change it back manually.

core/lib/jobs/jobs/jobs.py
```diff
@@ -219,28 +219,66 @@ def _get_placeholders_str(num_placeholders, start=1):
...
 
 
+async def get_customer_quota(con, customer_id, service) -> Dict[str, int]:
+    """Query DB and return usage limit and current usage.
+    Returns:
+        - Dictionary with account limits and account usage
+    """
+    # get service specific usage restrictions for the customer account
+    # uploads limit, jobs limit, end date of plan
+    usage_limit_query = "SELECT usage_restrictions->$1 AS usage FROM customers WHERE id=$2"
```
tannermpeterson, line 229: There is a lot of unnecessary data getting queried here. For example, `total_jobs` from `current_usage_query` isn't used at all. There's also 3 queries used to calculate the current usage, but that can be achieved with just one query:
```
SELECT COUNT(*) AS total_uploads, SUM(jobs_count) AS total_jobs
FROM (
    SELECT (
        CASE WHEN (COUNT(*) <= 2 AND COUNT(*) > 0) THEN 1 ELSE COUNT(*) END
    ) AS jobs_count FROM jobs_result
    WHERE customer_id=$1 AND type=$2
    GROUP BY upload_id
) dt;
```

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -477,7 +476,12 @@ async def create_new_job(
                     upload_file_to_s3(bucket=PULSE3D_UPLOADS_BUCKET, key=key, file=pv_parquet_path)
 
         return JobResponse(
-            id=job_id, user_id=user_id, upload_id=details.upload_id, status="pending", priority=priority
+            id=job_id,
```
luciipak, line 479: I also want to just double check about if you still need 443-445 (similar check also being performed in POST/uploads).. this was being used to display to users if limits had been reached or not, if you have another system set up, you might not need to return that error response? 

Following that, is it still set up that the 're-analysis existing recording' is disabled depending on if the credits are used up? We'll also want to disabled the interactive analysis option in the dropdown once it's reached as well.

deployments/pulse3d/services/pulse3d/src/main.py
tannermpeterson: Are we returning it in the route that creates a new job?

frontend/pulse3d/src/components/basicWidgets/UsageWidgetFull.js
tannermpeterson: Yeah, only the first re-analysis is free. We should change this

frontend/pulse3d/src/pages/upload-form.js
tannermpeterson: I think this should say something more along the lines of "Each successfully uploaded file will have analysis run on it, consuming 1 analysis credit each"

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
tannermpeterson: consume instead of use

frontend/pulse3d/src/pages/upload-form.js
tannermpeterson: consume instead of use

frontend/pulse3d/src/serviceWorker.js
tannermpeterson: I think we want to avoid changing info from the server in the FE like this. Also if this is just to help keep the usage up to date in the FE, I think a better way of doing this would be to poll the usage every 10 seconds or so (in case another user logged in under that customer account does something that changes the usage), and also hit the usage route any time the new job route is successfully hit

luciipak: I don't think we'd need to poll, its initially set up to return a response with an error key if a limit is reached in the POST routes, that should be sufficient for keeping track. That's what we need to make sure still works with these changes.

luciipak: and we want to make sure that when job limits are reached, that the re-analysis tab and interactive tab are disabled
which is what the error response above triggers

deployments/pulse3d/services/pulse3d/src/main.py
NikitaPoly: Yes. the job route does return this so I think it would not make sense to keep it.

deployments/pulse3d/services/pulse3d/src/main.py
NikitaPoly: The interactive analysis, new analysis and re-analysis all get disabled when usage limit gets met. I think it would be a good idea to leave the check. If the users find a way around the frontend and kick of a job manually, it will get prevented by this check. Would it be better to keep this for security ?

frontend/pulse3d/src/serviceWorker.js
NikitaPoly: The front end does get restricted when usage limit is reached and I have also removed this from this code from the service worker. There will be "lag" between job creation and displaying credit usage on the front end, but users are always notified if a credit was used. A check in both front end and back end prevent the running of any jobs above the limits.

frontend/pulse3d/src/components/basicWidgets/UsageWidgetFull.js
```diff
@@ -0,0 +1,51 @@
...
+          display: flex;
+          align-items: center;
+          color: black;
+          flex-flow: column;
+          text-align: center;
+        }
+        p#smallDescription {
+          font-size: 0.75rem;
+          text-align: center;
+        }
```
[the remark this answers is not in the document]
NikitaPoly: I was testing to see if jsx style tag would fix the inconsistent css styles on loading. It did not so I will change it back to be constant with the rest of the project. Although I do think readability and flexibility is improved with the jsx style tags.

core/db/curibio/alembic/versions/2ccc9fe21ec1_adding_plan_end_date_to_usage_.py
```diff
@@ -0,0 +1,28 @@
+"""Adding plan end date to usage_restrictions json object structure
+
+Revision ID: 2ccc9fe21ec1
+Revises: 1f23e047d2b0
```
tannermpeterson, line 4: Down revision should be changed here as well

core/lib/jobs/jobs/jobs.py
```diff
@@ -219,28 +219,58 @@ def _get_placeholders_str(num_placeholders, start=1):
...
+    # upload with 1 - 2 jobs  = 1 credit , upload with 3+ jobs = 1 credit for each upload with over 2 jobs
+    current_usage_query = "SELECT COUNT(*) AS total_uploads, SUM(jobs_count) AS total_jobs FROM ( SELECT ( CASE WHEN (COUNT(*) <= 2 AND COUNT(*) > 0) THEN 1 ELSE COUNT(*) END ) AS jobs_count FROM jobs_result WHERE customer_id=$1 AND type=$2 GROUP BY upload_id) dt"
+
+    async with con.transaction():
+        usage_limit_json = await con.fetchrow(usage_limit_query, service, customer_id)
+        current_usage_data = await con.fetchrow(current_usage_query, customer_id, service)
+
+    usage_limit_dict = json.loads(usage_limit_json["usage"])
+    usage_limit_dict["expiration_date"] = (
+        "" if usage_limit_dict["expiration_date"] is None else usage_limit_dict["expiration_date"]
```
tannermpeterson, line 240: Is there any reason we can't just leave this as `None`?

deployments/pulse3d/services/pulse3d/src/models/models.py
```diff
@@ -11,6 +11,13 @@ class UploadRequest(BaseModel):
...
```
tannermpeterson, line 14: UsageQuota*

frontend/pulse3d/src/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pulse",
-  "version": "0.3.3",
+  "version": "0.3.4",
```
luciipak, line 3: I had to bump prod to v0.3.4 in the release yesterday, could we make this 0.3.5?

frontend/pulse3d/src/serviceWorker.js
```diff
@@ -93,6 +93,9 @@ const clearAccountInfo = () => {
 const isLoginRequest = (url) => {
   return url.pathname.includes("/login");
 };
+const isNewUploadOrJob = (url, method) => {
```
luciipak, line 96: Can we remove this if it's no longer being used

core/lib/jobs/jobs/jobs.py
tannermpeterson: If you kept None it would just be `Dict[str, Optional[str]]` right?

core/db/curibio/alembic/versions/2ccc9fe21ec1_adding_plan_end_date_to_usage_.py
```diff
@@ -0,0 +1,28 @@
+"""Adding plan end date to usage_restrictions json object structure
+
+Revision ID: 2ccc9fe21ec1
+Revises: 89e819e0ecc2
```
luciipak, line 4: This needs to be changed now since another migration got merged before this one

luciipak (approved):

PR #364, opened 2023-01-19
alembic

NikitaPoly:


luciipak (approved):
Make sure to push the new worker image and run upgrade on the DB

PR #365, opened 2023-01-19
Pulse3d version bump

NikitaPoly:


core/db/curibio/alembic/versions/d7a72568d90b_add_pulse3d_version_0_29.1_0.py
```diff
@@ -0,0 +1,26 @@
+"""add pulse3d version 0.29.1
```
tannermpeterson, line 1: You can just change the upgrade script for 0.29.0 instead of adding this one. Also there is a typo in the name of this file

tannermpeterson (changes requested):

tannermpeterson:
Remember to delete the migration for 0.29.0

tannermpeterson (approved):

PR #371, opened 2023-01-24
email case insensitivity and peaks and valley indicator

luciipak:


tannermpeterson:
Should go into RC-02-02-23

PR #372, opened 2023-01-24
Case sensitivity

luciipak:


deployments/apiv2/services/users/src/main.py
```diff
@@ -324,7 +324,7 @@ async def register(
             )
 
             query_params = (
-                details.username.lower(),
+                details.username,
```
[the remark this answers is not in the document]
tannermpeterson: We should probably add it to wherever emails are used in the BE as well

core/db/curibio/alembic/versions/89e819e0ecc2_case_insensitive_emails.py
```diff
@@ -0,0 +1,24 @@
...
+Revises: d7a72568d90b
+Create Date: 2023-01-24 06:55:02.841578
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "89e819e0ecc2"
+down_revision = "d7a72568d90b"
```
tannermpeterson, line 13: We need to add one more revision into RC-01-19-23, so once that's added and merged into this branch we'll need to update the down revision here

tannermpeterson (changes requested):

core/db/curibio/alembic/versions/89e819e0ecc2_case_insensitive_emails.py
tannermpeterson: We have to wait until it's merged. I'll let you know what to change the down revision to. You won't need to rerun the migrations or anything

tannermpeterson: e04dfb9b43a8

core/db/curibio/alembic/versions/89e819e0ecc2_case_insensitive_emails.py
```diff
@@ -0,0 +1,24 @@
+"""case_insensitive_emails
+
+Revision ID: 89e819e0ecc2
+Revises: d7a72568d90b
```
tannermpeterson, line 4: e04dfb9b43a8

tannermpeterson (approved):

PR #378, opened 2023-01-31
Docker images run with non-root user

tannermpeterson:


deployments/apiv2/manifests/base/apiv2-dep.yaml
```diff
@@ -50,7 +50,7 @@ spec:
         ports:
         - containerPort: 9001
       - name: mantarray
-        image: 077346344852.dkr.ecr.us-east-2.amazonaws.com/mantarray:0.2.1
+        image: 077346344852.dkr.ecr.us-east-2.amazonaws.com/mantarray:0.3.0
```
luciipak, line 53: Do we also want to go ahead and move these to a test overlay like the pulse3d deployment?

NikitaPoly (approved):

PR #407, opened 2023-02-17
adding new node group, labels, namespace annotations 

jperson:


luciipak:
What will get defaulted to now? For example, the jobs-operator in the kopf namespace?

jperson:
> What will get defaulted to now? For example, the jobs-operator in the kopf namespace?

I'll have to add an annotation for that namespace, but if a namespace doesn't have an annotation it can use any node in the cluster

luciipak (dismissed):

tannermpeterson (approved):

PR #420, opened 2023-02-28
usage restriction fixes

NikitaPoly:


deployments/pulse3d/services/pulse3d/src/models/models.py
```diff
@@ -78,5 +78,5 @@ class UploadDownloadRequest(BaseModel):
 
 
 class GenericErrorResponse(BaseModel):
-    message: Union[str, Dict[str, bool]]
+    message: UsageQuota
```
luciipak, line 81: This needs to keep the string option,  `message:  Union[str, UsageQuota]`

luciipak: or `str | ...`

frontend/pulse3d/src/components/basicWidgets/CircularProgressWithLabel.js
```diff
@@ -4,7 +4,7 @@ import Typography from "@mui/material/Typography";
 export default function CircularProgressWithLabel(props) {
   return (
     <Box sx={{ position: "relative", display: "inline-flex" }}>
-      <CircularProgress variant="determinate" {...props} style={{ color: "var(--teal-green)" }} />
+      <CircularProgress variant="determinate" value={props.value} style={{ color: "var(--teal-green)" }} />
```
luciipak, line 7: why was this necessary?

luciipak:
Did you fix the issue with the credit usage warning modal not popping up with  interactive analysis?

frontend/pulse3d/src/components/basicWidgets/CircularProgressWithLabel.js
NikitaPoly: This removed an annoying error message from the console. It is not present in production but fires off every page load when testing.

luciipak: You can also set default props and do `CircularProgressWithLabel({ value, colorOfTextLabel = "black" })` so remove the extra code and conditional

luciipak:
These need to be merged into 02.16.23 also

PR #424, opened 2023-03-01
added polling for usage quota

NikitaPoly:


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -769,3 +768,17 @@ async def get_versions(request: Request):
     except Exception as e:
         logger.error(f"Failed to retrieve info of pulse3d versions: {repr(e)}")
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+@app.get("/usage-quota", response_model=Union[UsageQuota, GenericErrorResponse])
```
luciipak, line 773: I would just call this `/usage`

luciipak: the `GenericErrorResponse` never gets returned here so I don't think it's necessary as a `response_model`

luciipak: and if this exists now, i don't know if it's necessary to check in the jobs and uploads routes

NikitaPoly: I think this is still necessary because those are the checks that will prevent an analysis if a job gets through the FE for a user that has used all his quota

frontend/pulse3d/src/components/basicWidgets/UsageProgressWidget.js
```diff
@@ -20,52 +20,69 @@ const ExpiredP = styled.p`
...
-
+  const [isExpired, setIsExpired] = useState();
+  const pollUsageQuota = async () => {
+    try {
+      const response = await fetch(`${process.env.NEXT_PUBLIC_PULSE3D_URL}/usage-quota`);
+      if (response && response.status === 200) {
+        const newUsageQuota = await response.json();
+        const limit = parseInt(newUsageQuota.limits.jobs);
+        const actual = parseInt(newUsageQuota.current.jobs);
+        const usagePercentage = Math.round((actual / limit) * 100);
```
luciipak, line 35: since usagePercentage is already being used as a variable name in state, I would name this something else so there isn't any unexpected behaviors

frontend/pulse3d/src/components/basicWidgets/UsageProgressWidget.js
```diff
@@ -20,52 +20,69 @@ const ExpiredP = styled.p`
...
+
+  return (
+    <>
+      {maxJobs === -1 && <ProgressDiv>Unlimited Access</ProgressDiv>}
+      {!isExpired && maxJobs !== -1 && (
         <ProgressDiv>
           <p>Usage</p>
           <CircularProgressWithLabel value={usagePercentage} colorOfTextLabel={colorOfTextLabel} />
-          <ProgressP>{`${actualUploads}/${maxUploads} Analysis used`}</ProgressP>
+          <ProgressP>{`${actualJobs}/${maxJobs} Analysis used`}</ProgressP>
```
luciipak, line 82: I'd make this `Analyses`

deployments/pulse3d/services/pulse3d/src/main.py
luciipak: Eventually, we'll want to remove it from the jobs and uploads routes and just call this route before posting a new job etc and only call POST /jobs POST /uploads if this new route returns an unreached quota. But for now to get this out, we can just keep it.

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -769,3 +768,17 @@ async def get_versions(request: Request):
...
+async def get_usage_quota(request: Request, token=Depends(ProtectedAny(scope=PULSE3D_USER_SCOPES))):
```
luciipak, line 774: I also think you'll want to add customer scopes here so that it is displayed for admin accounts

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -769,3 +768,17 @@ async def get_versions(request: Request):
...
+    """Get the usage quota for the spesific user"""
```
luciipak, line 775: specific*

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -769,3 +768,17 @@ async def get_versions(request: Request):
...
+    try:
+        customer_id = str(uuid.UUID(token["customer_id"]))
```
luciipak, line 777: and for customer accounts, this will be the `userid` instead of `customer_id`, so once these changes are made, doubel check this works for both account types

frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -176,9 +176,11 @@ export default function UploadForm() {
         tabSelection === "Re-analyze Existing Upload" &&
         usageQuota &&
         usageQuota.limits &&
-        parseInt(usageQuota.limits.jobs) !== -1
+        parseInt(usageQuota.limits.jobs) !== -1 &&
+        files.length > 0 &&
+        files[0].created_at !== files[0].updated_at
     );
-  }, [paramErrors, files, inProgress, wellGroupErr]);
+  }, [paramErrors, files, inProgress]);
```
luciipak, line 183: `wellGroupErr` needs to be here

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -771,3 +770,23 @@ async def get_versions(request: Request):
     except Exception as e:
         logger.error(f"Failed to retrieve info of pulse3d versions: {repr(e)}")
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+@app.get("/usage", response_model=UsageQuota)
+async def get_usage_quota(
+    request: Request, token=Depends(ProtectedAny(scope=PULSE3D_SCOPES)), service: str = "pulse3d"
```
luciipak, line 777: this is an example how to use Query params 
https://github.com/CuriBio/k8s_curi/blob/4d110d238be88e7135cf96729d5adfd4d42cefca/deployments/pulse3d/services/pulse3d/src/main.py#L100

frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -176,7 +176,9 @@ export default function UploadForm() {
         tabSelection === "Re-analyze Existing Upload" &&
         usageQuota &&
         usageQuota.limits &&
```
luciipak, line 178: last nit picky comment, can we break up this big conditional into smaller ones with variable names that explain what it's checking? It's kind of hard to follow. Or at min, add some comments. How does this check if it's not the first reanalysis?

frontend/pulse3d/src/components/basicWidgets/UsageProgressWidget.js
```diff
@@ -20,52 +20,70 @@ const ExpiredP = styled.p`
...
-      usageState = (
-        <>
-          <ProgressDiv>Unlimited Access</ProgressDiv>
-        </>
-      );
-    } else if (!usageQuota.jobs_reached) {
-      // max usage has not been reached
-      usageState = (
+  useEffect(() => {
+    if (actualAnalyses) {
```
luciipak, line 67: is a new interval going to be started every time the actualAnalyses value changes?

NikitaPoly: I can confirm that the set interval is not being duplicated

NikitaPoly: anytime the interval is started the previous one is cancelled

luciipak (approved):

PR #431, opened 2023-03-13
Builder and rbac updates

luciipak:


bin/curi_builder/builder/__init__.py
```diff
@@ -80,7 +85,10 @@ def find_changed_svcs(sha: str):
 
     list_to_return = []
     for ch_path in changed_svc_paths:
-        dep_type, dep_name, *_, svc = ch_path.split("/")
+        if "queue-processor" not in ch_path:
+            dep_type, dep_name, *_, svc = ch_path.split("/")
+        else:
+            dep_name, svc = ch_path.split("/")
```
tannermpeterson, line 91: Won't line 100 error if dev_type isn't set?

luciipak: Can i swap the order and it will just pass after the first conditional?
`elif dep_type == "deployments" or svc in ("queue-processor", "jobs-operator")`
to `elif svc in ("queue-processor", "jobs-operator") or dep_type == "deployments"`

PR #432, opened 2023-03-13
updated builder image and rbac

luciipak:


bin/curi_builder/builder/__init__.py
```diff
@@ -80,7 +85,10 @@ def find_changed_svcs(sha: str):
 
     list_to_return = []
     for ch_path in changed_svc_paths:
-        dep_type, dep_name, *_, svc = ch_path.split("/")
+        if "queue-processor" not in ch_path:
+            dep_type, dep_name, *_, svc = ch_path.split("/")
+        else:
+            dep_name, svc = ch_path.split("/")
```
tannermpeterson, line 91: I think not setting dep_type will cause issues

tannermpeterson: Also would be a little more readable if the conditional was `if "queue-processor" in ch_path:`

tannermpeterson (approved):

PR #459, opened 2023-03-22
meta data cannot be null

NikitaPoly:


jperson (approved):

jobs/pulse3d/pulse3d-worker/src/lib/utils.py
```diff
@@ -33,13 +34,13 @@ def format_metadata(meta_sheet, pr, recording_length: int):
         "file_creation_timestamp": meta_sheet.iloc[11, 2],
         "mantarray_recording_session_id": uuid.uuid4(),
         "uploading_computer_name": well_file.get(COMPUTER_NAME_HASH_UUID, None),
-        "acquisition_started_at": well_file[UTC_BEGINNING_DATA_ACQUISTION_UUID],
-        "session_log_id": well_file.get(BACKEND_LOG_UUID, None),
+        "acquisition_started_at": well_file.get(UTC_BEGINNING_DATA_ACQUISTION_UUID,datetime.datetime.now()),
```
tannermpeterson, line 37: This value is the time that the data stream started on the physical instrument. Using `datetime.datetime.now()` wouldn't really be an acceptable replacement for this value. I think since it's not present in the metadata of the xlsx input file we should just set it to None instead

NikitaPoly: There is a DB constraint that does not allow for None type metadata. This would also error `                metadata["acquisition_started_at"].replace(tzinfo=None),` without a datetime object

tannermpeterson: We'll need to figure out a better value then, I think using `datetime.datetime.now()` could lead to confusion down the road

jperson: could have it as

```python
metadata["acquisition_started_at"].replace(tzinfo=None) if metadata["acquisition_started_at"] else None
```

PR #480, opened 2023-04-01
Add 'running' job status

luciipak:


core/db/curibio/alembic/versions/39405902fa31_add_running_to_status_enum_in_jobs_.py
```diff
@@ -0,0 +1,32 @@
...
+down_revision = "827f22860c04"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("""ALTER TYPE "JobStatus" ADD VALUE 'running'""")
+
+
+def downgrade():
```
tannermpeterson, line 22: I think you'll also want to change any jobs with the running status to pending

luciipak: Yeah better to ensure you're right

tannermpeterson (approved):

PR #486, opened 2023-04-11
Adjustable endpoints IA

NikitaPoly:


frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -15,6 +15,33 @@ import { AuthContext } from "@/pages/_app";
 
 const twentyFourPlateDefinition = new LabwareDefinition(4, 6);
 
+const wellNameToIndex = {
```
tannermpeterson, line 18: We can use `twentyFourPlateDefinition` defined above to do this

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -552,31 +571,148 @@ export default function WaveformGraph({
...
+        const y2 = y.invert(d3.select(this).attr("y2"));
+        const x2 = x.invert(
+          id.includes("peak")
+            ? d3.select("#peakLineY2Marker").attr("cx")
+            : d3.select("#valleyLineY2Marker").attr("cx")
+        );
+        const slope = id.includes("peak")
+          ? peakSlope[wellNameToIndex[currentWell]]
+          : valleySlope[wellNameToIndex[currentWell]];
+        const yintersept = y2 - slope * x2;
```
tannermpeterson, line 692: typo `yIntercept`

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -185,6 +212,16 @@ export default function InteractiveWaveformModal({
     startTime: null,
     endTime: null,
   });
+  //state for peaks
+  const [peakSlope, setPeakSlope] = useState(Array(24).fill(0));
```
tannermpeterson, line 216: Why are all 4 of these values needed for each? Can't we just store the two y values?

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -539,15 +585,22 @@ export default function InteractiveWaveformModal({
       // only filter if well data has been fetched, otherwise assume no filtering required because user would not have been able to have moved min peak and max valley lines
       if (well in originalData.coordinates) {
         const wellCoords = originalData.coordinates[well];
-        wellPeaks = wellPeaks.filter((peak) => wellCoords[peak][1] >= peakValleyWindows[well].minPeaks);
-        wellValleys = wellValleys.filter(
-          (valley) => wellCoords[valley][1] <= peakValleyWindows[well].maxValleys
-        );
+        wellPeaks = wellPeaks.filter((peak) => {
+          const actualY = wellCoords[peak][1];
+          const computedY =
```
tannermpeterson, line 590: I think actualY and computedY are confusing names. I'm assuming `actualY` is referring to the y-val of the peak/valley and `computedY` is referring the y-val of the filter line at that same x-val

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -454,9 +470,11 @@ export default function WaveformGraph({
       .data(
         initialPeaksValleys[0].filter((peak) => {
           const isPeakWithinWindow = dataToGraph[peak][0] >= startTime && dataToGraph[peak][0] <= endTime;
-          return peakValleyWindows[currentWell].minPeaks
-            ? isPeakWithinWindow && dataToGraph[peak][1] >= peakValleyWindows[currentWell].minPeaks
-            : isPeakWithinWindow;
+          const actualY = dataToGraph[peak][1];
+          const computedY =
+            dataToGraph[peak][0] * peakSlope[wellNameToIndex[currentWell]] +
```
tannermpeterson, line 475: I think peak slope should just be calculated as needed since we have enough information to do so and it's not an expensive calculation

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -552,31 +571,148 @@ export default function WaveformGraph({
...
+          d3.select("#valleyLine").attr(yId, yPosition);
+        } else {
+          d3.select("#peakLine").attr(yId, yPosition);
+        }
       })
-      .on("end", function () {
-        const id = d3.select(this).attr("id");
-        const yPosition = d3.select(this).attr("y1");
+      .on("end", function (d) {
+        // descrease stroke width when unselected and dropped
```
tannermpeterson, line 590: typo, I think it's supposed to be `decrease`

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -552,31 +571,148 @@ export default function WaveformGraph({
...
+        d3.select(this).attr("stroke-width", 2);
 
-        d3.select(this).attr("y1", yPosition).attr("y2", yPosition);
+        const id = d3.select(this).attr("id");
 
-        const pvCopy = peakValleyWindows;
-        if (id === "peakLine") {
-          pvCopy[currentWell].minPeaks = y.invert(yPosition);
-          setPeakValleyWindows({ ...pvCopy });
+        //need to invert to make calculation compatible with dataToGrpah
```
tannermpeterson, line 595: typo `dataToGraph`

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -552,31 +571,148 @@ export default function WaveformGraph({
...
+          id.includes("valley") ? d3.select("#valleyLine").attr("y2") : d3.select("#peakLine").attr("y2")
+        );
+
+        const x1 = x.invert(d3.select("#peakLineY1Marker").attr("cx"));
+        const x2 = x.invert(d3.select("#peakLineY2Marker").attr("cx"));
+
+        const slope = (y2 - y1) / (x2 - x1);
+        const yintersept = y2 - slope * x2;
+        // set variables used in marker deletion calculation
+        if (id.includes("peak")) {
```
tannermpeterson, line 609: Both branches of this if/else are very similar, there is probably a way to reduce the amount of duplicate code written here

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -552,31 +571,148 @@ export default function WaveformGraph({
...
+        //set starting y position
+        const yPosition = id.includes("valley")
+          ? Math.max(d.y, y(yMax + yRange))
+          : Math.min(d.y, y(yMin - yRange));
+        d3.select(this).attr("startingY", yPosition);
+      })
+      .on("drag", function (d) {
+        const id = d3.select(this).attr("id");
+
+        const yPosition = id.includes("valley")
```
tannermpeterson, line 657: This same code is written a few lines above, we should make a function to do this

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -552,31 +571,148 @@ export default function WaveformGraph({
...
+          ? Math.max(d.y, y(yMax + yRange))
+          : Math.min(d.y, y(yMin - yRange));
+        const initialY = d3.select(this).attr("startingY");
+        const changeInY = yPosition - initialY;
+        const newY1 = parseFloat(d3.select(this).attr("y1")) + changeInY;
+        const newY2 = parseFloat(d3.select(this).attr("y2")) + changeInY;
+
+        if (id.includes("peak")) {
```
tannermpeterson, line 665: ditto for refactoring to remove duplicate code

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -552,31 +571,148 @@ export default function WaveformGraph({
...
+          id.includes("peak")
+            ? d3.select("#peakLineY2Marker").attr("cx")
+            : d3.select("#valleyLineY2Marker").attr("cx")
+        );
+        const slope = id.includes("peak")
+          ? peakSlope[wellNameToIndex[currentWell]]
+          : valleySlope[wellNameToIndex[currentWell]];
+        const yintersept = y2 - slope * x2;
+
+        if (id.includes("peak")) {
```
tannermpeterson, line 694: This appears to be very similar to an if/else block I made a comment on further above. I think in addition to refactor the if/else branches to reduce duplicate code, we can also make this a function

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -883,7 +1117,33 @@ export default function WaveformGraph({
             left="5px"
             fontSize={15}
             borderRadius="5px"
-            clickFn={resetWellChanges}
+            clickFn={() => {
+              //set array here
+              let newArr = [...peakY1];
```
tannermpeterson, line 1122: This seems like the same 3 lines of code written over and over again, and I think I've seen these same 3 lines elsewhere. We should make a function for this as well

tannermpeterson (changes requested):
comments left individually

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
NikitaPoly: They are used for two different things. The one in the drag start event sets an initial y position to compare the dragging distance. The second one keeps getting the new position of where it is dragged to. Then they are compared to get the change in y

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
NikitaPoly: This code in necessary for the state monitoring in react to recognize that the array has changed. Without this the state will not be updated and the limit lines will not be moved.

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
NikitaPoly: I do think peak slope could be calculated, but I think storing it makes the code less confusing. I can make the change but I would like to leave it because it makes the code easier to read.

frontend/pulse3d/src/utils/labwareCalculations.js
```diff
@@ -100,4 +100,38 @@ export class WellTitle {
   getWellIndexFromRowColumn(rowIndex, columnIndex) {
     return columnIndex * this.numRows + rowIndex;
   }
+  /**
+   *
+   * @param {string} wellName
+   * @returns {int}
+   */
+  wellNameToIndex(wellName) {
+    const wellNameToIndexMap = {
```
tannermpeterson, line 109: This isn't what we want, it should basically be doing to opposite of whatever `wellIndexToName` is doing. We don't need to hardcode a mapping like this

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
tannermpeterson: Why does it make the code less confusing? I don't think the slope is necessary in any other parts of the code

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -552,31 +571,148 @@ export default function WaveformGraph({
...
+        );
+        const y2 = y.invert(
+          id.includes("valley") ? d3.select("#valleyLine").attr("y2") : d3.select("#peakLine").attr("y2")
+        );
+
+        const x1 = x.invert(d3.select("#peakLineY1Marker").attr("cx"));
+        const x2 = x.invert(d3.select("#peakLineY2Marker").attr("cx"));
+
+        const slope = (y2 - y1) / (x2 - x1);
+        const yintersept = y2 - slope * x2;
```
tannermpeterson: typo

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -431,6 +441,15 @@ export default function InteractiveWaveformModal({
...
     setPeakValleyWindows(existingData.peakValleyWindows);
+    setPeakSlope(existingData.peakSlope);
+    setPeakYIntercept(existingData.peakYIntercept);
+    setPeakY1(existingData.peakY1);
+    setPeakY2(existingData.peakY2);
+    setValleySlope(existingData.valleySlope);
+    setValleyYIntercept(existingData.valleyYIntercept);
+    setValleyY1(existingData.valleyY1);
+    setValleyY2(existingData.valleyY2);
+    getWaveformData(true, selectedWell);
```
luciipak, line 452: what happens in the changelog for this? and the "undo" feature?

luciipak: and the save feature?

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -143,8 +143,24 @@ export default function WaveformGraph({
...
+  peakY2,
+  setPeakY2,
+  valleySlope,
+  setValleySlope,
+  valleyYIntercept,
+  setValleyYIntercept,
+  valleyY1,
+  setValleyY1,
+  valleyY2,
+  setValleyY2,
```
luciipak, line 163: this seems like way too many props to pass down? is there a way to condense these?

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
tannermpeterson: The first place I would start (as mentioned in one of my other comments) is making a function for this same 3 lines of code that is repeated over and over here:
```
const updateArray = (updateFn, originalArray, newVal) => {
  let newArr = [...originalArray];
  newArr[wellIndex] = newVal; // could also parametrize wellIndex
  updateFn(newArr);
};
```
Now we can swap in the function:
```
if (id.includes("peak")) {
  updateArray(setPeakY1, peakY1, y1);
  updateArray(setPeakY2, peakY2, y2);
  updateArray(setPeakYIntercept, peakYIntercept, yIntercept);
  updateArray(setPeakSlope, peakSlope, slope);
} else {
  updateArray(setValleyY1, valleyY1, y1);
  updateArray(setValleyY2, valleyY2, y2);
  updateArray(setValleyYIntercept, valleyYIntercept, yIntercept);
  updateArray(setValleySlope, valleySlope, slope);
}
```
At this point another pattern becomes apparent, `updateArray` is being called over and over. There are few different ways reduce down to one line of code actually calling `updateArray`, here's just one:
```
const updateArgs = id.includes("peak")
      ? [
          [setPeakY1, peakY1, y1],
          [setPeakY2, peakY2, y2],
          [setPeakYIntercept, peakYIntercept, yIntercept],
          [setPeakSlope, peakSlope, slope],
        ]
      : [
          [setValleyY1, valleyY1, y1],
          [setValleyY2, valleyY2, y2],
          [setValleyYIntercept, valleyYIntercept, yIntercept],
          [setValleySlope, valleySlope, slope],
        ];
    for (const args of updateArgs) updateArray(...args);
```
You could argue readability was reduced by this change however, so I think it's fine to stick with just the changes I mentioned prior to this

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -585,29 +721,81 @@ export default function WaveformGraph({
...
+      .attr("id", "peakLineY1Marker")
+      .attr("lineId", "peakLine")
+      .attr("r", 6)
+      .attr("cx", x((endTime - startTime) / 100))
+      .attr("cy", y(peakY1[wellNameToIndex[currentWell]]))
+      .attr("fill", "orange")
+      .attr("stroke", "orange")
+      .style("cursor", "pointer")
+      .call(pivotLineDrag);
+    const peaksY2 = svg
```
tannermpeterson, line 718: I think we could make a function for this as well

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
tannermpeterson: The `updateArray` function I wrote in another comment can be used here

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
NikitaPoly: A function to create each draggable point at the ends of each line?

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
NikitaPoly: The save feature saves all the slopes, intercepts, y1 and y2 for each line in each well. On load a modal shows up to confirm loading from last session or start over.

The undo feature only undoes changes to wells. I need to add the limit lines to the change log and undo feature.

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
tannermpeterson: Whatever this is doing. There is a lot of duplicate code here creating `peaksY1` and `peaksY2`. I think the same is true for the valley equivalents

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -552,31 +584,80 @@ export default function WaveformGraph({
...
+    const moveLineUpDown = d3
+      .drag()
+      .on("start", function (d) {
+        const id = d3.select(this).attr("id");
+        // close context menu if it's open
+        contextMenu.style("display", "none");
+        // increase stroke width when selected and dragging
+        d3.select(this).attr("stroke-width", 5);
+        //set starting y position
+        const initialY = id.includes("valley")
```
tannermpeterson, line 628: Seeing this a lot as well, should make a function for this to

frontend/pulse3d/src/components/interactiveAnalysis/WaveformGraph.js
```diff
@@ -732,11 +850,36 @@ export default function WaveformGraph({
...
     startTime,
     endTime,
-    peakValleyWindows,
   ]);
 
+  //run once when component is created to set starter values for calculations
+  useEffect(() => {
+    const well_index = twentyFourPlateDefinition.wellNameToIndex(currentWell);
+    if (!peakY1[well_index] || !peakY2[well_index]) {
+      let newArr = [...peakY1];
```
tannermpeterson, line 866: Should still make a function for these 3 lines

frontend/pulse3d/src/utils/labwareCalculations.js
```diff
@@ -100,4 +100,17 @@ export class WellTitle {
...
+  /**
+   * Get the well index from well name
+   *
+   * @param {string} wellName
+   * @returns {int}
+   */
+  wellNameToIndex(wellName) {
+    const row = wellName.charCodeAt(0) - "A".charCodeAt(0);
+    const col = parseInt(wellName.slice(1)) - 1;
+    const numCols = 6;
```
tannermpeterson, line 112: num_cols is already defined in the constructor on line 27 https://github.com/CuriBio/k8s_curi/pull/486/files#diff-f070e8d7209e1b23b3985652cefa077c2a575132a1cd7ebf420d3183f84b5bbaR27. We should use that value instead of redefining. Also, we should make that value set in the constructor `numCols` instead of `num_cols`

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -562,33 +584,69 @@ export default function InteractiveWaveformModal({
...
+        xRange.max,
+        {
+          minPeaks: findLowestPeak(selectedWell),
+          maxValleys: findHighestValley(selectedWell),
+        },
+        peakValleyWindows[selectedWell].maxValleys,
+        peakValleyWindows[selectedWell].maxValleys,
+        peakValleyWindows[selectedWell].minPeaks,
+        peakValleyWindows[selectedWell].minPeaks
+      );
```
luciipak, line 631: ` 
maxValleyY = findHighestValley(selectedWell);
minPeakY = findLowestPeak(selectedWell);

changelogMessage = getChangelogMessage(
        ogWellData[0],
        ogWellData[1],
        xRange.min,
        xRange.max,
        {
          minPeaks: minPeakY,
          maxValleys: maxValleyY,
        },
        maxValleyY,
        maxValleyY,
        minPeakY,
        minPeakY
      );
`

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -643,6 +705,30 @@ export default function InteractiveWaveformModal({
       changelogMessage = `Maximum valleys window changed from ${pvWindow.maxValleys.toFixed(
         2
       )} to ${peakValleyWindows[selectedWell].maxValleys.toFixed(2)}`;
+    } else if (isNewValleyY1 && isNewValleyY2) {
+      changelogMessage = `Valley Line moved ${
+        valleyY1[twentyFourPlateDefinition.getIndexFromWellName(selectedWell)] - valleyY1ToCompare
```
luciipak, line 710: `twentyFourPlateDefinition.getIndexFromWellName(selectedWell)` is reused a lot, I think it could be separated out somewhere above the if/else

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -700,6 +786,10 @@ export default function InteractiveWaveformModal({
       endTime,
       message,
       pvWindow: pvWindowCopy[selectedWell],
+      valleyYOne: valleyY1[twentyFourPlateDefinition.getIndexFromWellName(selectedWell)],
```
luciipak, line 789: ditto on comment above `twentyFourPlateDefinition.getIndexFromWellName(selectedWell)`

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -826,9 +973,20 @@ export default function InteractiveWaveformModal({
...
+            peakY1={peakY1}
+            setPeakY1={setPeakY1}
+            peakY2={peakY2}
+            setPeakY2={setPeakY2}
+            valleyY1={valleyY1}
+            setValleyY1={setValleyY1}
+            valleyY2={valleyY2}
+            setValleyY2={setValleyY2}
+            calculateYLimit={calculateYLimit}
+            twentyFourPlateDefinition={twentyFourPlateDefinition}
```
luciipak, line 987: I think this would be better imported into WaveformGraph the same way as above instead of being passed as props

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
luciipak: Actually, I think this should just be set somewhere in state if it's being called so much. Just set selectedWellIdx as a state value and then only need to call this once whenever the selected well changes.

frontend/pulse3d/src/utils/labwareCalculations.js
```diff
@@ -20,11 +20,11 @@ export class WellTitle {
    * Take pixel coordinates from a drawing and convert it back to the x/y numerical values that should have been used to generate those pixel coordinates.
    *
    * @param {int} numRows - The number of rows in the labware/plate
-   * @param {int} num_columns - The number of columns in the labware/plate
+   * @param {int} NumCols - The number of columns in the labware/plate
```
luciipak, line 23: this needs to be camel case

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -562,33 +580,80 @@ export default function InteractiveWaveformModal({
...
+        valleys,
+        startTime,
+        endTime,
+        pvWindow,
+        valleyYOne,
+        valleyYTwo,
+        peakYOne,
+        peakYTwo,
+      } = wellChanges[wellChanges.length - 1];
+      changelogMessage = getChangelogMessage(
```
tannermpeterson, line 609: With this many parameters to a function I think it makes sense to just pass in an object instead of each indivodual value

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -791,6 +887,55 @@ export default function InteractiveWaveformModal({
...
+    setValleyY1(newArr);
+    newArr = [...valleyY2];
+    newArr[wellIdx] = peakValleyWindows[selectedWell].maxValleys;
+    setValleyY2(newArr);
+  };
+
+  const setBothLinesToNew = (newPeakY1, newPeakY2, newValleyY1, newValleyY2) => {
+    let arr = [...peakY1];
+    arr[wellIdx] = newPeakY1;
+    setPeakY1(arr);
```
tannermpeterson, line 924: Is there a reason we can make a function for these 3 lines repeated everywhere?

NikitaPoly: I think we might need to leave these specific statements because when added a function like this
```
const assingNewArr = (data, newValue) => {
    let newArr = [...data];
    newArr[wellIdx] = newValue;
    return newArr;
  };
```
The state stops being updated. The reason a new array needs to be created and assigned is because react does not react to changes unless the length of the array changes or a completely new Array is passed.

tannermpeterson: I posted in an earlier comment how the function would need to be written https://github.com/CuriBio/k8s_curi/pull/486#discussion_r1169462647

NikitaPoly: I also tried 

```
function assingNewArr(data, newValue, setState) {
    let newArr = [...data];
    newArr[wellIdx] = newValue;
    setState(newArr);
  }
```
I did not update state either

luciipak: changing `setState(newArr);` to `setState([...newArr]);` should change state

tannermpeterson (approved):

PR #494, opened 2023-04-18
update Loki target revision branch

luciipak:


NikitaPoly (approved):

cluster/manifests/grafana-loki/base/grafana-loki.yaml
```diff
@@ -6,15 +6,15 @@ metadata:
...
-  - repoURL: https://grafana.github.io/helm-charts
-    chart: loki-stack
-    targetRevision: 2.9.9
-    helm:
-      valueFiles:
-      - $values/cluster/manifests/grafana-loki
-  - repoURL: https://github.com/CuriBio/k8s_curi.git
-    targetRevision: RC-03-30-23 # TODO must change before merging to main
-    ref: values
+    - repoURL: https://grafana.github.io/helm-charts
```
tannermpeterson, line 9: I think these indentations can sometimes cause problems. If your formatter added them automatically you might need to set it to the "advanced" option for yaml or something like that

PR #500, opened 2023-04-25
Pulse3D 0.32.4

tannermpeterson:


core/db/curibio/alembic/versions/3ffac4d68585_pulse3d_0_32_4.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "3ffac4d68585"
+down_revision = "39405902fa31"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("INSERT INTO pulse3d_versions (version, state) VALUES ('0.32.4', 'external')")
```
luciipak, line 19: do we also want to deprecate the old version here? 0.32.2

tannermpeterson: I think it's fine to handle deprecations manually. Also still not sure if we need to keep 0.32.2 for any reason yet

luciipak (approved):

PR #503, opened 2023-04-27
swap pulse3d 0.32.4 for 0.32.5

luciipak:


jperson (dismissed):

deployments/pulse3d/services/pulse3d/src/requirements.txt
```diff
@@ -4,7 +4,7 @@ click==8.0.3
 fastapi==0.70.1
 h11==0.12.0
 idna==3.3
-pulse3d==0.32.4
+pulse3d==0.32.5
```
tannermpeterson, line 7: Think we still need to do this in the worker as well

luciipak: yeah that was weird, I just replaced searched and replaced all 0.32.4, idk why it missed that one

tannermpeterson (approved):

PR #518, opened 2023-05-08
Pulse3d dropdown selection when filtered xlsx versions are present

luciipak:


NikitaPoly (approved):

tannermpeterson (approved):

frontend/pulse3d/src/components/uploadForm/AnalysisParamForm.js
```diff
@@ -197,7 +202,15 @@ export default function AnalysisParamForm({
     });
 
     setPulse3dVersionOptions([...options]);
-  }, [pulse3dVersions, metaPulse3dVersions, xlsxFilePresent]);
+  }, [pulse3dFilteredFileVersions, metaPulse3dVersions]);
+
+  useEffect(() => {
```
tannermpeterson, line 207: What happens if you choose a pulse3d version that doesn't support xlsx and then upload and xlsx file? Also what happens to the custom params that are entered in that situation?

luciipak: if `xlsxFilePresent` is false, then the second conditional will be true and essentially no pulse3d versions will be filtered out

luciipak: custom params will be sent as is.. but maybe I'm missing what you're asking.. I don't think anything will be different is `xlsxFilePresent` is false

PR #522, opened 2023-05-10
IA bug fixes

NikitaPoly:
IA Bug fixes

fixed bug with extra change log entrees : added check for null values
fixed bug where switching wells after changing window would cause windows to flicker between positions : refactored extra state changes
fixed but where filtering peaks an valleys on submit would use the last selected well data for all wells.


luciipak (approved):

frontend/pulse3d/src/components/interactiveAnalysis/InteractiveAnalysisModal.js
```diff
@@ -271,20 +271,17 @@ export default function InteractiveWaveformModal({
...
-      const dataToCompare = originalData.coordinates[wellToUse];
-
-      peaksList = peaksList.filter((peak) => dataToCompare[peak][1] >= peakValleyWindows[wellToUse].minPeaks);
-      valleysList = valleysList.filter(
-        (valley) => dataToCompare[valley][1] <= peakValleyWindows[wellToUse].maxValleys
-      );
-    }
-
+    // Filter before looking for duplicates
+    const wellCoords = originalData.coordinates[wellToUse]
```
tannermpeterson, line 280: In general for JS, if you see the pattern `c = a ? a : b;` you can reduce it to `c = a || b;`

tannermpeterson (approved):

PR #523, opened 2023-05-11
New analysis params

tannermpeterson:


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -127,8 +129,8 @@ async def get_info_of_uploads(
 
             return uploads
 
-    except Exception as e:
```
NikitaPoly, line 130: Why do we not log the error anymore?

tannermpeterson: We do, see https://docs.python.org/3/library/logging.html#logging.Logger.exception

NikitaPoly (approved):

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -393,13 +395,25 @@ async def create_new_job(
...
 
+        if use_noise_based_peak_finding:
+            params += [
+                "height_factor",
+                "relative_prominence_factor",
+                "noise_prominence_factor",
+                "max_frequency",
+                "valley_search_duration",
+                "upslope_duration",
+                "upslope_noise_allowance_duration",
```
luciipak, line 412: I keep meaning to ask about if we can all go over these new params soon.

deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -616,7 +632,7 @@ def _yield_s3_objects(bucket: str, keys: List[str], filenames: List[str]):
             yield filenames[idx], datetime.now(), 0o600, ZIP_64, obj.get()["Body"]
 
     except Exception as e:
-        raise S3Error(f"Failed to access {bucket}/{key}: {repr(e)}")
+        raise S3Error(f"Failed to access {bucket}/{key}") from e
```
luciipak, line 635: what does `from e` do?

frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -276,37 +300,70 @@ export default function UploadForm() {
...
 
+      for (const [name, value] of [
+        ["normalize_y_axis", normalizeYAxis],
+        ["twitch_widths", twitchWidths],
+        ["start_time", startTime],
+        ["end_time", endTime],
+        ["max_y", maxY],
+        ["include_stim_protocols", showStimSheet],
+      ]) {
+        requestBody[name] = getNullIfEmpty(value);
```
luciipak, line 316: 👍

luciipak (approved):

deployments/pulse3d/services/pulse3d/src/main.py
tannermpeterson: In the traceback it makes it clear that a new exception was intentionally raised vs. an additional exception getting raised while trying to handle the original. Here's an example traceback I found with both:
```
raceback (most recent call last):
  File "asyncio\locks.py", line 214, in wait
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "asyncio\tasks.py", line 456, in wait_for
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "controller\subsystems\instrument_comm.py", line 350, in _handle_beacon_tracking
  File "asyncio\tasks.py", line 458, in wait_for
asyncio.exceptions.TimeoutError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "controller\utils\aio.py", line 39, in clean_up_tasks
  File "controller\subsystems\instrument_comm.py", line 355, in _handle_beacon_tracking
controller.exceptions.SerialCommStatusBeaconTimeoutError
```
where `The above exception was the direct cause of the following exception:` means that `from e` was used and `During handling of the above exception, another exception occurred:` means it wasn't

deployments/pulse3d/services/pulse3d/src/main.py
tannermpeterson: Yeah we should, it's actually not too complicated

PR #533, opened 2023-05-25
changed filtering in handleNewAnalysis

NikitaPoly:
Jobs were not being posted for re-analyze option because `includes` was not matching a file object when checking if file data already exists. The check now iterates over all file id's instead of whole object when checking if file data is already present.

### DIFF
--- a/frontend/pulse3d/src/pages/upload-form.js
+++ b/frontend/pulse3d/src/pages/upload-form.js
@@ -265,7 +265,6 @@ export default function UploadForm() {
   };
 
   const postNewJob = async (uploadId, filename) => {
-    console.log("posting job");
     try {
       const {
         normalizeYAxis,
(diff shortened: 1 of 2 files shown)


frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -455,9 +455,12 @@ export default function UploadForm() {
       setInProgress(true);
 
       for (const file of files) {
+        //check file is in uploads list
+        const fileIsInList = uploads.some((upload) => upload.id === file.id);
         if (file instanceof File) {
           await uploadFile(file);
-        } else if (uploads.includes(file)) {
+        } else if (fileIsInList) {
+          console.log("posting job");
```
tannermpeterson, line 463: console.log

tannermpeterson (approved):

PR #540, opened 2023-05-30
test file structure

NikitaPoly:
moved to (https://github.com/CuriBio/k8s_curi/pull/549 )because git was adding files that were not changed


jperson:
I think the tests directory should be at the top level of the `frontend` dir or even at the top level of the repo, since they don't depend on anything in `frontend` or `frontend/pulse3d` and eventually we are going to support more products than pulse3d in this repo and I think it would be better to have a single directory that all the tests get put under.

I'd also prefer to use python instead of javascript for the playwright tests, and please remove the test code that was copied from some other repo

PR #543, opened 2023-05-31
bug fix

NikitaPoly:
Fixed bug where changing parameter would disable the submit button. 

- selected file to reanalyze gets cleared when submitting a job or navigating away from the page 
- submit button disable now checks for file to reanalyze as well

### DIFF
--- a/frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
+++ b/frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
@@ -22,7 +22,7 @@ export default function InputDropdownWidget({
       setSelected(options[defaultIndex]);
       handleSelection(defaultIndex);
     }
-  }, [options]);
+  }, [defaultIndex]);
 
   return (
     <Autocomplete
--- a/frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
+++ b/frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
@@ -11,12 +11,10 @@ export default function InputDropdownWidget({
   handleSelection,
   defaultIndex,
 }) {
-  const [selected, setSelected] = useState(options[defaultIndex] ? options[defaultIndex] : null);
+  const [selected, setSelected] = useState(options[defaultIndex] || null);
 
   useEffect(() => {
-    if (reset) {
-      setSelected(null);
-    }
+    if (reset) setSelected(null);
   }, [reset]);
 
   useEffect(() => {
--- a/frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
+++ b/frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
@@ -16,7 +16,6 @@ export default function InputDropdownWidget({
   useEffect(() => {
     if (reset) {
       setSelected(null);
-      console.log("here");
     }
   }, [reset]);
 
--- a/frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
+++ b/frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
@@ -9,7 +9,7 @@ export default function InputDropdownWidget({
   label,
   reset,
   handleSelection,
-  defaultFile,
+  defaultFileIndex,
 }) {
   const [selected, setSelected] = useState(null);
 
@@ -18,12 +18,12 @@ export default function InputDropdownWidget({
   }, [reset]);
 
   useEffect(() => {
-    // Set default file name if one is present
-    if (defaultFile && defaultFile.length !== 0) {
-      setSelected(defaultFile);
-      handleSelection(options.indexOf(defaultFile));
+    if (defaultFileIndex && defaultFileIndex !== -1) {
+      console.log(defaultFileIndex);
+      setSelected(options[defaultFileIndex]);
+      handleSelection(defaultFileIndex);
     }
-  }, [defaultFile]);
+  }, [options]);
 
   return (
     <Autocomplete
--- a/frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
+++ b/frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
@@ -11,16 +11,17 @@ export default function InputDropdownWidget({
   handleSelection,
   defaultIndex,
 }) {
-  const [selected, setSelected] = useState(options[defaultIndex]);
+  const [selected, setSelected] = useState(options[defaultIndex] ? options[defaultIndex] : null);
 
   useEffect(() => {
-    if (reset) setSelected(null);
+    if (reset) {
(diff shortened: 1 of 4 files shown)


frontend/pulse3d/src/pages/_app.js
```diff
@@ -117,6 +117,12 @@ function Pulse({ Component, pageProps }) {
     return () => clearInterval(swInterval);
   }, [router.pathname]);
 
+  useEffect(() => {
+    //clear default file when user leaves the re-analyze page
+    if (router.query.id !== "Re-analyze Existing Upload") {
+      setDefaultReanalysisFile(null);
+    }
+  }, [router.query]);
```
luciipak, line 125: Can this go in the uploads form component itself? Or does it need to be in this top level comp?

frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -175,7 +175,11 @@ export default function UploadForm() {
     // checks if error value exists, no file is selected, or upload is in progress
     const checkConditions =
       !Object.values(paramErrors).every((val) => val.length === 0) ||
-      !((files.length > 0 && files[0] instanceof File) || (uploads && uploads.includes(files[0]))) ||
+      !(
+        (files.length > 0 && files[0] instanceof File) ||
+        (uploads && uploads.includes(files[0])) ||
+        defaultReanalysisFile
```
tannermpeterson, line 181: We should try to unify the hooks for `files` and `defaultReanalysisFile` here. We're starting to go down the same road we did with IA where we're keeping two copies of the same state and it's causing problems. We can just extract `const [files, setFiles] = useState([]);` up to the parent component and set that directly instead of using this new `defaultReanalysisFile`. @luciipak does this seem like a better solution?

luciipak: Yeah I agree that there are already mechanisms set up to reset both the drag and drop zone and the input dropdown that would be easier to use than adding another state value. I'm not as up to speed on the bug, but the component should handle resetting itself. 
Additionally, I noticed that you're passing a `defaultFile` value down to the InputDropdownWidget, it's best to not add component specific props to a reusable component. I can tell that you're trying to set the default selected value to a specific file name... it would be better to add some initialValue  index prop instead so that other components can use it the same way.  

```
 useEffect(() => {
    // Set default file name if one is [present](url)
    if (defaultFile && defaultFile.length !== 0) {
      setSelected(defaultFile);
      handleSelection(options.indexOf(defaultFile));
    }
  }, [defaultFile]);
```

should be something like:
```
 useEffect(() => {
    if (initialValue) {
      setSelected(options[initialValue]);
      handleSelection(initialValue);
    }
  }, [initialValue]);
```

frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
```diff
@@ -9,7 +9,7 @@ export default function InputDropdownWidget({
   label,
   reset,
   handleSelection,
-  defaultFile,
+  defaultFileIndex,
```
luciipak, line 12: should remove use of the word file in general to make it reusable to components without file names as the options

frontend/pulse3d/src/pages/upload-form.js
luciipak: I don't think the comments above were addressed in the PR changes.. removing use of defaultReanalysisFile and using existing mechanisms for resetting state

frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
```diff
@@ -18,12 +18,11 @@ export default function InputDropdownWidget({
...
-    // Set default file name if one is present
-    if (defaultFile && defaultFile.length !== 0) {
-      setSelected(defaultFile);
-      handleSelection(options.indexOf(defaultFile));
+    if (defaultIndex && defaultIndex !== -1) {
+      setSelected(options[defaultIndex]);
+      handleSelection(defaultIndex);
     }
-  }, [defaultFile]);
+  }, [options]);
```
luciipak, line 25: this should be dependent on the defaultIndex?

frontend/pulse3d/src/pages/upload-form.js
NikitaPoly: I am using the given mechanisms to reset state, but there are none for resetting state when the page is navigated away from. The use effect in the `_app` is for resetting state of default file so that it is not present when the user attempts to do a normal reanalysis. I'm also not sure how I would remove `defaultReanalysisFile` without adding the selected file as a query param to the URL. 

The state for selected file is saved in the _app context and passed to children components who need it. When a user submits a reanalysis all built in mechanisms are used to reset state. But when you navigate away from the reanalysis page  without submitting the job and back, the default file still get set because it was never cleared in `_app.js`.  Is there a better way of keeping the state or passing the selected file for reanalysis ?

frontend/pulse3d/src/components/basicWidgets/InputDropdownWidget.js
```diff
@@ -9,21 +9,19 @@ export default function InputDropdownWidget({
   label,
   reset,
   handleSelection,
-  defaultFile,
+  defaultIndex,
 }) {
-  const [selected, setSelected] = useState(null);
+  const [selected, setSelected] = useState(options[defaultIndex] ? options[defaultIndex] : null);
```
tannermpeterson, line 14: `options[defaultIndex] || null`

frontend/pulse3d/src/pages/upload-form.js
```diff
@@ -599,9 +611,9 @@ export default function UploadForm() {
               options={formattedUploads}
               width={500}
               label="Select Recording"
-              reset={files.length === 0}
+              reset={resetFilesDropDown && files.length === 0}
               handleSelection={handleDropDownSelect}
-              defaultFile={defaultReanalysisFile}
+              defaultIndex={formattedUploads.indexOf(defaultReanalysisFile)}
```
tannermpeterson, line 616: It would probably be better to just pass in the default option directly here

frontend/pulse3d/src/pages/upload-form.js
tannermpeterson: My original point was that we know have `defaultReanalysisFile`, `setDefaultReanalysisFile`, and `const [files, setFiles] = useState([]);`. It's fine to have all these, but the default reanalysis file should only be used when initializing the component, i.e. something like this `const [files, setFiles] = useState(defaultReanalysisFile ? [defaultReanalysisFile] : []);`. That way it is used more appropriately as an initial value and not a duplicate copy of state that floats around in this component

frontend/pulse3d/src/pages/upload-form.js
NikitaPoly: I would still need the index to call `handleSelection`. So that would be adding in extra prop. Would it be ok to keep it as is?

PR #570, opened 2023-06-21
Product-specific QP environment variables

luciipak:


jobs/pulse3d/manifests/base/job-runner.yaml
```diff
@@ -7,5 +7,9 @@ spec:
   job_queue: pulse3d
   seconds_to_poll_db: 15
   max_num_of_workers: 8
+  ecr_repo: null
+  product_specific:
+    pulse3d_uploads_bucket: null
+    mantarray_logs_bucket: null
```
luciipak, line 13: Does anyone have an opinion if they think I should capitalize all of these env variables?

jobs/jobs-operator/manifests/base/job-runner-crd.yaml
```diff
@@ -25,6 +25,9 @@ spec:
                   type: integer
                 seconds_to_poll_db:
                   type: integer
+                product_specific:
+                  type: object
+                  x-kubernetes-preserve-unknown-fields: true
```
luciipak, line 30: Does anyone have an opinion if they think I should capitalize all of these env variables in this CRD?

tannermpeterson: What would that look like?

jobs/queue-processor/src/main.py
```diff
@@ -25,9 +25,12 @@
...
 
-# TODO figure out how to pass these
-MANTARRAY_LOGS_BUCKET_ENV = kclient.V1EnvVar(name="MANTARRAY_LOGS_BUCKET_ENV", value="prod-mantarray-logs")
-UPLOADS_BUCKET_ENV = kclient.V1EnvVar(name="UPLOADS_BUCKET_ENV", value="prod-pulse3d-uploads")
+MANTARRAY_LOGS_BUCKET = kclient.V1EnvVar(
+    name="MANTARRAY_LOGS_BUCKET_ENV", value=os.getenv("mantarray_logs_bucket")
+)
+PULSE3D_UPLOADS_BUCKET = kclient.V1EnvVar(
+    name="UPLOADS_BUCKET_ENV", value=os.getenv("pulse3d_uploads_bucket")
+)
```
luciipak, line 33: @tannermpeterson it would just mean these would be capitalized in `os.getenv`

tannermpeterson: I think to be consistent let's capitalize

luciipak: *we decided not to since the other CRDs do not

jobs/pulse3d/manifests/base/job-runner.yaml
```diff
@@ -7,5 +7,9 @@ spec:
   job_queue: pulse3d
   seconds_to_poll_db: 15
   max_num_of_workers: 8
+  ecr_repo: null
+  product_specific:
+    pulse3d_uploads_bucket: null
```
jperson, line 12: Does this need to be product specific? Since each product will have it's own s3 bucket for data/logs it seems like this could be `uploads_bucket` and `analyzed_bucket` and the queue processor will grab those and pass them to the jobs

luciipak: @jperson  Yeah we can do that. I didn't know what future products would need so was just leaving it open ended to add whichever and however many, but happy to change it if you think we'll always need the same variables for each product.

jperson (approved):

PR #571, opened 2023-06-21
fixed pre-commit errors

NikitaPoly:


luciipak (approved):

e2e/pulse3d/login/test_admin_login.py
```diff
@@ -1,12 +1,12 @@
 import pytest
 
-from fixtures import basic_page, setup, video_setup
+from fixtures import basic_page, setup, video_setup  # noqa: F401
```
tannermpeterson, line 3: In general it's good to attempt to try and fix the issue instead of just suppressing it. I'm pretty sure you can fix this by doing `from .fixtures import`. Even if that doesn't fix this pre-commit issue, we should be doing that because that is how it works in all of our other test suites

PR #573, opened 2023-06-23
Waveform snapshots

luciipak:


frontend/pulse3d/src/utils/generic.js
```diff
@@ -101,6 +102,14 @@ const getWaveformCoordsFromTable = async (table, normalizeYAxis) => {
   return coordinatesObj;
 };
 
+const getTableFromParquet = async (buffer) => {
+  const wasmModule = await import("parquet-wasm/esm/arrow1.js");
```
tannermpeterson, line 106: Is it possible to import this module just once? Could you move this out into the global scope and import?

tannermpeterson (approved):

frontend/pulse3d/src/utils/generic.js
luciipak: No that causes issues, needs to be awaited in react

PR #581, opened 2023-06-27
More playwright tests

NikitaPoly:


luciipak:
you'll need to update the base branch. and just double checking that pre-commit has been run?

PR #619, opened 2023-08-09
Limit failed login attempts

luciipak:


core/db/curibio/alembic/versions/316e4d5eb991_add_failed_pass_attemps_column.py
```diff
@@ -0,0 +1,28 @@
+"""add failed pass attemps column
```
tannermpeterson, line 1: typo attempts

core/db/curibio/alembic/versions/4c5b40dd7e96_customer_account_alias.py
```diff
@@ -0,0 +1,22 @@
+"""customer account alias
```
tannermpeterson, line 1: How did this file get into this PR? I haven't merged it yet

deployments/apiv2/services/users/src/main.py
```diff
@@ -96,35 +96,46 @@ async def login(request: Request, details: UserLogin | CustomerLogin):
...
-        select_query = "SELECT password, id, data->'scope' AS scope FROM users WHERE deleted_at IS NULL AND name=$1 AND customer_id=$2 AND suspended='f' AND verified='t'"
+        select_query = (
+            "SELECT password, id, data->'scope' AS scope, failed_login_attempts, suspended FROM users "
+            "WHERE deleted_at IS NULL AND name=$1 AND customer_id=$2 AND verified='t'"
+        )
         select_query_params = (username, str(details.customer_id))
         customer_id = details.customer_id
 
-        update_last_login_query = "UPDATE users SET last_login = $1 WHERE deleted_at IS NULL AND name = $2 AND customer_id=$3 AND suspended='f' AND verified='t'"
+        update_last_login_query = "UPDATE users SET last_login = $1, failed_login_attempts = 0 WHERE deleted_at IS NULL AND name = $2 AND customer_id=$3 AND verified='t'"
```
tannermpeterson, line 119: nitpick, can we keep the spacing around `=` consistent, this query has both `x=y` and `x = y`

deployments/apiv2/services/users/src/main.py
```diff
@@ -198,6 +222,17 @@ async def login(request: Request, details: UserLogin | CustomerLogin):
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
 
+async def _update_failed_login_attempts(con, account_type: str, id: str, count: int) -> None:
+    if count == MAX_FAILED_LOGIN_ATTEMPTS:
+        # if max failed attempts is reached, then deactivate the account and increment count
+        update_query = f"UPDATE {account_type}s SET suspended='t', failed_login_attempts=failed_login_attempts+1 where id=$1"
```
tannermpeterson, line 228: Do we need to suspend the user? I think this works fine, but just wondering if we want to also suspend the user

luciipak: I think so because then the admin can reactive them using the mechanisms we already have set up

tannermpeterson: Good idea

tannermpeterson (approved):

PR #628, opened 2023-08-29
disallow login after max failed

luciipak:


deployments/apiv2/services/users/src/main.py
```diff
@@ -186,6 +186,10 @@ async def login(request: Request, details: UserLogin | CustomerLogin):
                 ph.hash(pw)
                 raise LoginError(failed_msg)
             else:
+                # only raise LoginError when account is locked on successful creds after they have been checked to prevent giving away facts about successful login combinations
+                if select_query_result["failed_login_attempts"] >= MAX_FAILED_LOGIN_ATTEMPTS:
```
tannermpeterson, line 190: I think we need to change how suspended accounts are handled as well. Right now if the account is suspended, it sets the query result to None which causes an error when this line is run `ph.verify(select_query_result["password"], pw)`. I think what we should do is essentially what you've just done but also check if the user is suspended:
```
if select_query_result["suspended"]:
   if select_query_result["failed_login_attempts"] >= MAX_FAILED_LOGIN_ATTEMPTS:
      raise LoginError(account_locked_msg)
   raise LoginError(failed_msg)
```
and remove the suspended check prior to running `ph.verify`

luciipak: yeah that error was happening because the successful login wasn't resetting the suspended status, that's prevented now

tannermpeterson: We should also update the unit tests

tannermpeterson: If the account is suspended for some other reason though the error will still occur

luciipak: true

tannermpeterson (approved):

PR #646, opened 2023-09-22
fix for user registration and preset bug

luciipak:


jperson (approved):

deployments/apiv2/services/users/src/main.py
```diff
@@ -366,7 +367,7 @@ async def logout(request: Request, token=Depends(ProtectedAny(check_scope=False)
 
 @app.post("/register", response_model=UserProfile | CustomerProfile, status_code=status.HTTP_201_CREATED)
 async def register(
-    request: Request, details: CustomerCreate | UserCreate, token=Depends(ProtectedAny(scope=CUSTOMER_SCOPES))
```
tannermpeterson, line 369: I think there might have been a reason they were ordered this way, but can't remember

luciipak: I'm not sure, but this is definitely what was causing the issue

PR #650, opened 2023-09-30
Updates to /login

luciipak:


deployments/apiv2/services/users/src/main.py
```diff
@@ -78,142 +78,121 @@ async def startup():
...
+            return LoginResponse(tokens=tokens, usage_quota=usage_quota, user_scopes=avail_user_scopes)
+
+    except LoginError as e:
+        raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED, detail=str(e))
+    except Exception:
+        logger.exception("POST /login/customer: Unexpected error")
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+@app.post("/login", response_model=LoginResponse)
```
tannermpeterson, line 132: Did we decide not to also add `/login/user`? We could just do 
```
@app.post("/login/user", response_model=LoginResponse)
async def _login_user(request: Request, details: UserLogin):
   return await loging_user(request, details)
```
Or something like that so they both work

PR #656, opened 2023-10-04
Case insensitive aliases

luciipak:


deployments/apiv2/services/users/src/main.py
```diff
@@ -153,7 +153,7 @@ async def login_user(request: Request, details: UserLogin):
         select_query = (
             "SELECT u.password, u.id, u.failed_login_attempts, u.suspended, u.customer_id "
             "FROM users AS u JOIN customers AS c ON u.customer_id=c.id "
-            "WHERE u.deleted_at IS NULL AND u.name=$1 AND c.alias=$2 AND u.verified='t'"
+            "WHERE u.deleted_at IS NULL AND u.name=$1 AND LOWER(c.alias)=LOWER($2) AND u.verified='t'"
```
tannermpeterson, line 156: Should we make sure that the query that adds an alias also converts it to lowercase?

luciipak: That's what i was asking about in slack, Jason was saying it would be fine just when checking

tannermpeterson: I think then we'd still be able to add `CuriBio` and `curibio` for example, and then we'd probably get multiple rows returned from this query. I know that's an unlikely scenario, but still might be worth addressing it

luciipak: Yeah then do we want to just auto correct uppercase to lowercase in the UI when entering so that if they save with uppercases, it doesn't look different when they login next and look at the alias? 
Also was there anything setup to check if an alias already exists? we wouldn't want that anyway

tannermpeterson: There should be a unique constraint on the row. I think if we convert to lowercase in the query that updates the alias it should be fine, might also need to run a DB migration to convert all aliases to lowercase. I think we had to do all this when we made the usernames case-insensitive as well

luciipak: What gets returned if the unique constraint fails? Does the user know why it fails?

tannermpeterson: I'm not actually sure what it shows, but I would assume that the route will return a 500

tannermpeterson (approved):

PR #658, opened 2023-10-04
Curi admin scope

luciipak:


core/db/curibio/alembic/versions/7b704d670049_add_curi_scope.py
```diff
@@ -0,0 +1,24 @@
...
+# revision identifiers, used by Alembic.
+revision = "7b704d670049"
+down_revision = "36891004f3e6"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        "INSERT INTO account_scopes VALUES ((SELECT id FROM customers WHERE email='software@curibio.com'), NULL, 'curi:admin')"
```
luciipak, line 19: not actually sure if this was necessary to do, just thought might as well

tannermpeterson: @jperson any thoughts on whether we should do this in a migration or manually?

jperson (approved):

PR #662, opened 2023-10-06
Add new customer accounts

luciipak:


deployments/apiv2/services/users/src/main.py
```diff
@@ -520,25 +541,30 @@ async def register_user(
...
+    """Send or resend account emails.
 
     No token required for request. Currently sending reset password and new registration emails based on query type.
     """
     # EmailRegistrationError will be raised if random type param is added that isn't verify or reset
     try:
         async with request.state.pgpool.acquire() as con:
             query = (
-                "SELECT id, customer_id, name FROM users WHERE email=$1"
+                "SELECT id, customer_id, name FROM users WHERE email=$1 AND id=$2"
```
tannermpeterson, line 559: Why do we need to check both the email and ID? Couldn't we just check one or the other?

tannermpeterson (approved):

deployments/apiv2/services/users/src/main.py
luciipak: It's to check that the email the user enters matches the customer id in jwt token sent in the initial email. We don't want them entering the wrong email or someone elses email

PR #670, opened 2023-10-19
fix mantarray svc

tannermpeterson:


luciipak:
why is it getting bumped two minor versions?

tannermpeterson:
I brought it in from another branch, guess it should've only got bumped by 1

PR #679, opened 2023-10-24
Pulse3d-worker logging

luciipak:
Exception example

Screenshot 2023-10-17 at 13 47 07

Example with bound ids
Screenshot 2023-10-25 at 07 37 10


jobs/pulse3d/pulse3d-worker/src/main.py
```diff
@@ -158,7 +163,6 @@ async def process(con, item):
                 logger.info("Starting pulse3d analysis")
                 if use_existing_time_v_force:
                     # if any plate recording args are provided, can't load from data frame since a re-analysis is required to recalculate the waveforms
-                    logger.info(f"Loading previous time force data from {parquet_filename}")
```
tannermpeterson, line 161: Is there any new log line replacing this one? Or is it fine to just get rid of this and not replace

luciipak: I was trying to reduce some of the logging and a previous step logs if time force data was found, I think it would be fine without, but don't feel strongly either way.

tannermpeterson (approved):

PR #699, opened 2023-12-01
Pulse3d rewrite integration

luciipak:


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -600,8 +612,9 @@ def _format_tuple_param(
         default_values = (default_values,) * len(options)
 
     # set any unspecified values to the default value
+    # pulse3d does not like float values
```
tannermpeterson, line 615: We could also have pulse3d enforce these values being int if needed, make it the responsibility of the code using the value instead of the caller

jobs/pulse3d/test-pulse3d-worker/src/main.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+# needs to be prefixed so that the queue processor doesn't pick it up
+@get_item(queue="test-pulse3d-v1.0.0rc9")
+async def process_item(con, item):
+    # keeping initial log without bound variables
+    logger.info(f"Processing item: {item}")
+
+    s3_client = boto3.client("s3")
+    job_metadata = {"processed_by": "1.0.0rc9"}
```
tannermpeterson, line 48: Could do `{"processed_by": PULSE3D_VERSION}`

jobs/pulse3d/test-pulse3d-worker/src/main.py
```diff
@@ -0,0 +1,360 @@
...
+
+                tissue_waveforms = pl.read_parquet(tissue_waveforms_path)
+                stim_waveforms = (
+                    pl.read_parquet(stim_waveforms_path) if os.path.exists(stim_waveforms_path) else None
+                )
+
+                metadata_dict = json.load(open(metadata_path))
+                is_beta_2 = metadata_dict["file_format_version"] >= VersionInfo.parse("1.0.0")
+                existing_metadata = (
+                    MantarrayBeta2Metadata(**metadata_dict)
```
tannermpeterson, line 127: Do we have support for nautilus files yet?

jobs/pulse3d/test-pulse3d-worker/src/main.py
```diff
@@ -0,0 +1,360 @@
...
+                except UnableToConvergeError:
+                    raise Exception("Unable to converge due to low quality of data")
+                except Exception:
+                    logger.exception("PreAnalysis failed")
+                    raise
+
+            try:
+                # copy so that windowed data isn't written to S3 and used on following recordings
+                windowed_pre_analyzed_data = deepcopy(pre_analyzed_data)
+                # TODO could window stim data as well?
```
tannermpeterson, line 165: Yeah we'll probably need to window that as well, might be nice to have pulse3d handle all the windowing so the worker doesn't need to worry about it if we need to start windowing other things in the future

jobs/pulse3d/test-pulse3d-worker/src/main.py
```diff
@@ -0,0 +1,360 @@
...
+            try:
+                outfile_prefix = prefix.replace("uploads/", "analyzed/test-pulse3d/")
+                outfile_key = f"{outfile_prefix}/{job_id}/{output_filename}"
+                upload_file_to_s3(bucket=PULSE3D_UPLOADS_BUCKET, key=outfile_key, file=output_filename)
+                logger.info(f"Uploaded {output_filename} to {PULSE3D_UPLOADS_BUCKET}/{outfile_key}")
+            except Exception:
+                logger.exception("Upload failed")
+                raise
+
+            if not interactive_analysis:
```
tannermpeterson, line 247: We might want to upload all this intermediate data right after it's done so that if a downstream module errors we don't skip uploading to S3

jobs/pulse3d/test-pulse3d-worker/src/main.py
tannermpeterson, line 1: Looking a lot cleaner overall 👍

jobs/pulse3d/test-pulse3d-worker/src/main.py
luciipak: yeah I like that

jobs/pulse3d/test-pulse3d-worker/src/main.py
luciipak: yes, true

luciipak: @tannermpeterson does the new way make sense?

deployments/pulse3d/services/pulse3d/src/main.py
luciipak: Yeah I'll make a note to also add it to pulse3d

jperson (approved):

jobs/pulse3d/test-pulse3d-worker/src/main.py
```diff
@@ -38,14 +45,27 @@
 PULSE3D_VERSION = "v1.0.0rc9"
 
 
+def _get_existing_metadata(
+    metadata_dict: dict[str, Any]
+) -> BaseMetadata | MantarrayBeta2Metadata | MantarrayBeta1Metadata:
```
tannermpeterson, line 50: This type can just be BaseMetadata since the other two are subclasses of it

jobs/pulse3d/test-pulse3d-worker/src/main.py
tannermpeterson: Yeah I guess since we don't have a Nautilus metadata class we'll need to use BaseMetadata for now. Eventually we could make a function that takes the instrument type in the metadata dict or something and spits out the correct metadata class

tannermpeterson (approved):

PR #700, opened 2023-12-04
Firmware uploading

tannermpeterson:


deployments/apiv2/services/mantarray/src/main.py
```diff
@@ -124,55 +142,220 @@ async def delete_serial_number(
...
     except Exception:
         err_msg = f"{fw_type.title()} Firmware v{version} not found"
         logger.exception(err_msg)
         return JSONResponse(status_code=status.HTTP_404_NOT_FOUND, content={"message": err_msg})
+
+
+@app.post("/firmware/{fw_type}/{version}")
+async def upload_firmware_file(
+    request: Request,
+    details: MainFirmwareUploadRequest | ChannelFirmwareUploadRequest,
```
luciipak, line 253: What's the difference again between main firmware and channel firmware?

frontend/pulse3d/src/pages/production-console.js
```diff
@@ -1,22 +1,725 @@
...
+  return (
+    <TableContainer>
+      <TableHeader>Serial Numbers</TableHeader>
       <SerialNumberTable />
-    </Container>
+    </TableContainer>
+  );
+}
+
+export default function ProductionConsole() {
```
luciipak, line 708: Do these all need to go on the same page or can we split them out so it's a bit easier to follow?

frontend/pulse3d/src/pages/production-console.js
```diff
@@ -1,22 +1,725 @@
...
+  };
+
+  const getFileName = (file) => {
+    return file ? file.name : null;
+  };
+
+  const noErrors = () => {
+    return Object.values(uploadOptionErrors).every((val) => val.length === 0);
+  };
+  fwTypes.includes(uploadOptions.fwType);
```
luciipak, line 211: is this meant to be here?

deployments/apiv2/services/mantarray/src/main.py
tannermpeterson: The main FW runs only on the main microcontroller of the instrument, the channel FW runs on each of the microcontrollers under each well

frontend/pulse3d/src/pages/production-console.js
tannermpeterson: These are probably fine all being on the same page

frontend/pulse3d/src/pages/production-console.js
tannermpeterson: Not sure, looks like it's not doing anything though

frontend/pulse3d/src/pages/production-console.js
luciipak: I mean all the components, can we break them up so it's not such a large file?

tannermpeterson: Yeah I can split it up

tannermpeterson: Actually I think that might cause problems with the styling. I can make a note to break everything up once we find a way to reuse all the styling

luciipak (approved):

PR #706, opened 2023-12-19
Structlog dep fix

luciipak:
`We` were using threadlocal module in structlog (https://www.structlog.org/en/23.2.0/thread-local.html) to manage binding of relevant request variables (eg customer id, user id, upload/job ids, ....) to the logger, which is now deprecated in favor of context module (https://www.structlog.org/en/stable/contextvars.html).

With this change as is, context variables bound in the request handler are not available to middleware handler which needs them. (see example https://github.com/tiangolo/fastapi/issues/4696).
There is also now a relevant warning in the structlog docs itself for fastapi:

```
Since the storage mechanics of your context variables is different for each concurrency method, they are isolated from each other.

This can be a problem in hybrid applications like those based on [starlette](https://www.starlette.io/) (this [includes FastAPI](https://github.com/tiangolo/fastapi/discussions/5999)) where context variables set in a synchronous context don’t appear in logs from an async context and vice versa.
```

The fix in this pr uses `startlette-context` to manage context through entire request-response cycle to store relevant request variables for use in the middleware logs and exceptions.

### DIFF
--- a/frontend/pulse3d/src/components/basicWidgets/ButtonWidget.js
+++ b/frontend/pulse3d/src/components/basicWidgets/ButtonWidget.js
@@ -1,0 +1,66 @@
+import styled from "styled-components";
+import CircularSpinner from "./CircularSpinner";
+const Button = styled.button(
+  ({ props }) => `
+  background-color: ${props.backgroundColor || "var(--dark-blue)"};
+  color: ${props.color || "var(--light-gray)"};
+  font-size: inherit;
+  border: none;
+  width: 100%;
+  height: 100%;
+  position: ${props.position || "inherit"};
+  border-radius: ${props.borderRadius || "0px"};
+  font-size: ${props.fontSize || "18px"};
+  box-shadow: 3px 3px 3px 0px rgb(0 0 0 / 30%), 7px 7px 7px 0px rgb(0 0 0 / 10%);
+  &:hover {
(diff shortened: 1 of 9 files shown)


frontend/pulse3d/src/components/basicWidgets/ButtonWidget.js
```diff
@@ -0,0 +1,66 @@
...
+  position: relative;
+  width: ${props.width || "100%"};
+  height: ${props.height || "60px"};
+  top: ${props.top || "0px"};
+  left: ${props.left || "0px"};
+`
+);
+
+// TODO should this automatically disable itself if props.inProgress === true ?
+const ButtonWidget = (props) => {
```
tannermpeterson, line 46: What is this widget for?

luciipak: somewhere this file got removed and needs to be capitalized, do you have it as buttonWidget?

tannermpeterson: I'm not sure

luciipak: Well regardless it'll need to be added back to RC-12-07-23

tannermpeterson (approved):

PR #708, opened 2024-01-02
Customer to admin

tannermpeterson:


luciipak (commented):
Do we want to bump any versions?

tannermpeterson:
Wouldn't hurt but I know that pretty much every service has had it's version bumped in this RC branch already

luciipak (approved):

PR #710, opened 2024-01-04
New pulse3d into svc 

tannermpeterson:


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -536,22 +536,41 @@ async def create_new_job(
...
+            version = details.version
+            if version == "1.0.0":
+                version = "1.0.0rc13"
+
+            job_meta = {"analysis_params": analysis_params, "version": version}
             # if a name is present, then add to metadata of job
-            job_meta = {"analysis_params": analysis_params, "version": details.version}
             if details.name_override and pulse3d_semver >= "0.32.2":
-                job_meta.update({"name_override": details.name_override})
+                job_meta["name_override"] = details.name_override
```
luciipak, line 547: I'm just curious what the benefit of this is?

luciipak (approved):

deployments/pulse3d/services/pulse3d/src/main.py
tannermpeterson: They both do the same thing, but I think this way is a little more readable if you are adding a single entry to the dict

PR #725, opened 2024-01-13
Reset pw

luciipak:


deployments/apiv2/services/users/src/main.py
```diff
@@ -633,9 +650,16 @@ async def _create_account_email(
         if ScopeTags.ACCOUNT not in scope.tags:
             raise Exception(f"Scope {scope} is not allowed in an email token")
 
-        account_type = AccountTypes.USER if "user" in scope else AccountTypes.ADMIN
+        if "user" in scope:
+            account_type = AccountTypes.USER
+            account_id = user_id
```
tannermpeterson, line 655: Does `token.account_id` not work here?

deployments/apiv2/services/users/src/main.py
```diff
@@ -347,6 +347,23 @@ async def _update_failed_login_attempts(con, account_type: str, id: str, count:
...
+        try:
+            ph.verify(prev_pw, pw)
+        except VerifyMismatchError:
+            # passwords don't match, nothing else to do
+            continue
+        else:
+            # passwords match, return msg indicating that this is the case
+            raise UnableToUpdateAccountError()
+
+    await con.execute(update_query, *[phash, *query_params])
```
tannermpeterson, line 364: Minor note for future, I think `await con.execute(update_query, phash, *query_params)` should work fine here

tannermpeterson (approved):

deployments/apiv2/services/users/src/main.py
luciipak: oh duh yeah, I actually meant to change that

PR #727, opened 2024-01-23
Nautilai normalization

tannermpeterson:


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -813,7 +815,7 @@ async def get_interactive_waveform_data(
             logger.exception(message)
             return GenericErrorResponse(error="MissingDataError", message=message)
 
-        data_type_str: str | None = analysis_params.get("data_type")
+        data_type_str: str | None = parsed_meta.get("data_type")
```
luciipak, line 818: was this a bug then?

luciipak (approved):

deployments/pulse3d/services/pulse3d/src/main.py
tannermpeterson: It was a bug, it would only grab the correct data type if you set an override value

PR #730, opened 2024-01-24
fix s3 permissions

tannermpeterson:


luciipak (approved):
this will probabyl need to be manually added to the iam policy in aws, i don't think terraform registers these changes

PR #735, opened 2024-01-24
fix default p3d

tannermpeterson:


luciipak:
Nice was just about to work on this

luciipak (approved):
is this up in modl? I noticed that switching between reanalysis form and new upload form it doesn't set the params to match the version earlier, so just want to make sure that's working too now

tannermpeterson:
Not up in modl yet

PR #738, opened 2024-01-30
Admin edit customers

luciipak:


deployments/apiv2/services/users/src/main.py
```diff
@@ -804,7 +804,7 @@ async def get_all_users(request: Request, token=Depends(ProtectedAny(tag=ScopeTa
     query = (
         "SELECT u.id, u.name, u.email, u.created_at, u.last_login, u.verified, u.suspended, u.reset_token, array_agg(s.scope) as scopes "
         "FROM users u "
-        "INNER JOIN account_scopes s ON u.id=s.user_id "
+        "LEFT JOIN account_scopes s ON u.id=s.user_id "
```
luciipak, line 807: Switching to left join so that if for some reason a user (or customer below, see similar query) has no scopes in account_scopes, it still returns the user/customer entry with scopes as [] and doesn't just omit it

deployments/apiv2/services/users/src/main.py
```diff
@@ -982,6 +1013,83 @@ async def get_user(
...
+            async with request.state.pgpool.acquire() as con:
+                async with con.transaction():
+                    # get current customer scopes
+                    rows = await con.fetch(
+                        "SELECT scope FROM account_scopes WHERE customer_id=$1 AND user_id IS NULL",
+                        account_id,
+                    )
+
+                    # get product name from admin scopes
+                    existing_products = [dict(p)["scope"].split(":")[0] for p in rows]
```
tannermpeterson, line 1064: We have a function `get_product_tags_of_admin` that handles this

tannermpeterson (approved):

deployments/apiv2/services/users/src/main.py
luciipak: But the returned value from the fetch query would still need to be iterated through and the scopes in the db are strings so `s.tags` fails in `get_product_tags_of_admin`. Additionally, it needs to be compared against the list of strings sent in the request body

tannermpeterson: I think something like `get_product_tags_of_admin([convert_scope_str(row["scope"]) for row in rows])` should work. `Scopes` inherits from `StrEnum` so all the fields have all the string behavior. We also have `ScopeConverter` which our models can inherit from which also converts the strings to `Scopes`

tannermpeterson: This current method assumes that the first component of each scope is a product name which could change if we add a scope in the future which does not start with a product name

PR #800, opened 2024-05-23
Allow creation of admins with login_type = sso

roberto-araiza-curibio:
Hey guys,

I'm adding the ability for curi:admin to create admins with the selected login_type

Please let me know what you guys think. Thanks!

new-admin-login-type

### DIFF
--- a/deployments/apiv2/services/users/src/main.py
+++ b/deployments/apiv2/services/users/src/main.py
@@ -72,6 +72,7 @@ async def lifespan(app: FastAPI):
 
 app = FastAPI(openapi_url=None, lifespan=lifespan)
 
+LOGIN_TYPES = ["password", "sso_microsoft"]
 MAX_FAILED_LOGIN_ATTEMPTS = 10
 TEMPLATES = Jinja2Templates(directory="templates")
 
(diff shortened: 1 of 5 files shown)


jperson (approved):

deployments/apiv2/services/users/src/main.py
```diff
@@ -72,6 +72,7 @@ async def lifespan(app: FastAPI):
 
 app = FastAPI(openapi_url=None, lifespan=lifespan)
 
+LOGIN_TYPES = ["password", "sso_microsoft"]
```
tannermpeterson, line 75: nitpick, A `StrEnum` works nicely for this, it might also allow pydantic+fastAPI to check the values passed in to the route for us:
```
from enum import StrEnum, auto
class LoginTypes(StrEnum):
    PASSWORD = auto()
    SSO_MICROSOFT = auto()

...

class AdminCreate(ScopeConverter):
    ...
    login_type: LoginType = Field(default=LoginTypes.PASSWORD)
```

tannermpeterson (approved):

deployments/apiv2/services/users/src/main.py
roberto-araiza-curibio: this is way better.  I changed it.

PR #801, opened 2024-05-29
Allow admins to 'Sign in with Microsoft'

roberto-araiza-curibio:
Hey guys,

I'm adding a 'Sign in with Microsoft' option to our Admin sign-in page.

Please let me know if you have any questions, comments, or suggestions.

Thanks!
Roberto

Screenshot 2024-05-29 at 8 42 03 AM

Screenshot 2024-05-27 at 3 00 59 PM

---
  - https://app.asana.com/0/0/1207435942363891


deployments/apiv2/manifests/overlays/modl/kustomization.yaml
```diff
@@ -27,6 +27,11 @@ patches:
         value:
           name: DASHBOARD_URL
           value: https://dashboard.curibio-modl.com
+      - op: add
+        path: /spec/template/spec/containers/0/env/0  # should this be /spec/template/spec/containers/0/env/1 ??
```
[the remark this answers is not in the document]
jperson: you can also test this by running `kustomize build <path to overlays folder> > output.yaml` and then reviewing the output.yaml file

deployments/apiv2/manifests/overlays/modl/kustomization.yaml
```diff
@@ -27,6 +27,11 @@ patches:
...
+        value:
+          name: MICROSOFT_SSO_APP_ID
+          value: "e32a941e-eb86-4588-823a-1b3fbab61565"
```
jperson, line 34: is it ok for this to be public or should it be kept in a sealed secret?

tannermpeterson: I was thinking about that too. It looks like it's needed in the FE as well, so if we do need this kept secret it might be difficult getting into the FE

roberto-araiza-curibio: all the docs I've read state that the app id is public, e.g.:

"Client ID and secret – In this pattern, a pair of values is generated by the authorization server when registering a client. The client ID is a public value that identifies the application, while the client secret is a confidential value used to prove the identity of the application." [[link]](https://learn.microsoft.com/en-us/entra/identity-platform/msal-client-applications)

We don't use a secret since our front-end code is "... a form of public client application in which the client code is executed in a user-agent such as a web browser. Clients such as these don't store secrets because the browser context is openly accessible." [[link]](https://learn.microsoft.com/en-us/entra/identity-platform/msal-js-initializing-client-applications)

The solution for public client applications used to be the "implicit flow", now superseded by the [Authorization Code Flow with PKCE](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-pkce), which is implemented by MSAL ([docs](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-browser)).

Further, looking at the MSAL Initialization [docs](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/initialization.md) we see:

```
clientId: "ENTER_CLIENT_ID"
```

without the usual "don't commit this" warnings, so I think we're good here.

deployments/apiv2/manifests/overlays/modl/kustomization.yaml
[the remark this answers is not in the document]
roberto-araiza-curibio: I ran `kubectl kustomize overlays/prod` with both `.../env/0` (left in image below) and `.../env/1` and they both work:

Screenshot 2024-05-30 at 8 24 39 AM

I think that last number in `path` indicates the index to insert at

tannermpeterson (approved):

PR #814, opened 2024-06-10
Presets by product

tannermpeterson:
Pre-commit formatter ran on a number of files


deployments/apiv2/services/users/src/main.py
```diff
@@ -732,7 +748,7 @@ async def register_user(
                         email=email,
                         url=f"{DASHBOARD_URL}",
                         subject="Your User account has been created",
-                        template="registration_sso.html"
+                        template="registration_sso.html",
```
roberto-araiza-curibio, line 751: is there a linting step that I'm missing?

deployments/apiv2/services/users/src/models/users.py
```diff
@@ -1,10 +1,9 @@
 import re
-from enum import StrEnum, auto
 from typing import Any
 from uuid import UUID
 from pydantic import BaseModel, EmailStr, SecretStr, Field
 from pydantic import constr, field_validator
-from auth import AuthTokens, Scopes, ScopeConverter
+from auth import AuthTokens, Scopes, ScopeConverter, LoginType
```
roberto-araiza-curibio, line 6: thanks for fixing this, it was on my to-do list

roberto-araiza-curibio (approved):

deployments/apiv2/services/users/src/main.py
tannermpeterson: I think we forgot to mention how to set up pre-commit, it will handle linting, formatting, etc for us each time we commit. I don't remember the pre-commit set up steps exactly, but I think you can just pip install pre-commit globally and the once in the root dir of any repo run `pre-commit install` then every `git commit` will also run pre-commit. The .pre-commit-config.yaml files in our repos have the min pre-commit versions required

PR #855, opened 2024-08-02
P3d worker store metrics output

tannermpeterson:
Also:
- fix p3d output upload location
- remove metadata from pg_notify calls


jobs/queue-processor/src/main.py
tannermpeterson, line 1: Formatter must've run on this file, there shouldn't be any significant changes

roberto-araiza-curibio (approved):

PR #856, opened 2024-08-06
 Extract metrics

tannermpeterson:


bin/extract_metrics/install.sh
```diff
@@ -0,0 +1,4 @@
+export AWS_ACCOUNT_ID=725604423866
```
roberto-araiza-curibio, line 1: this AWS_ACCOUNT_ID is hard-coded because we only need it to install pulse?  am I reading that correctly?

bin/extract_metrics/extract_metrics.py
```diff
@@ -0,0 +1,339 @@
...
+import structlog
+from utils.s3 import upload_file_to_s3
+
+PULSE3D_UPLOADS_BUCKET = os.getenv("TEST_UPLOADS_BUCKET")
+
+DRY_RUN = True
+
+s3_client = boto3.client("s3")
+
+logging.config.dictConfig(
```
roberto-araiza-curibio, line 35: why the two different loggers?  does it merit a comment here stating the reason?

bin/extract_metrics/extract_metrics.py
```diff
@@ -0,0 +1,339 @@
...
+                logger.info(f"found {len(all_jobs)} jobs")
+                for job_info in all_jobs:
+                    res = await process_job(con, job_info)
+                    job_counts[res] += 1
+    finally:
+        logger.info(f"result: {job_counts}")
+        logger.info("DONE")
+
+
+def _create_file_info(
```
roberto-araiza-curibio, line 97: this is the same function we have in the pulse3d-worker, but I'm not sure it merits a refactor so we can reuse it if we're only running this backfill once

bin/extract_metrics/extract_metrics.py
```diff
@@ -0,0 +1,339 @@
...
+        DB_USER = os.getenv("POSTGRES_USER", default="curibio_jobs")
+        DB_HOST = os.getenv("POSTGRES_SERVER", default="psql-rds.default")
+        DB_NAME = os.getenv("POSTGRES_DB", default="curibio")
+
+        dsn = f"postgresql://{DB_USER}:{DB_PASS}@{DB_HOST}:5432/{DB_NAME}"
+
+        async with asyncpg.create_pool(dsn=dsn) as pool:
+            async with pool.acquire() as con:
+                all_jobs = await con.fetch(
+                    "select * from jobs_result where meta->>'version' like '1.0%' and meta->>'version' not like '%rc%' and status='finished' order by created_at desc"
```
roberto-araiza-curibio, line 85: does this query get all jobs with >= 1.0.0, and we're only guaranteed < 1.0.8 because we know no job in the DB has >= 1.0.8 ?

roberto-araiza-curibio (commented):
LGTM.  I have a few comments, but nothing major.

bin/extract_metrics/extract_metrics.py
tannermpeterson: p3d uses structlog which unfortunately means we have to go through this complicated setup to get logs written to both stdout and a file. I can add a comment with links explaining this

bin/extract_metrics/install.sh
tannermpeterson: We could allow this to be an environment variable, I just hard-coded it for convenience. You can install p3d from either the prod or modl cluster but not the test cluster, so I decided to be explicit

bin/extract_metrics/extract_metrics.py
tannermpeterson: Yes, however for the hermes project we do probably want to move this into a shared library at some point

bin/extract_metrics/extract_metrics.py
tannermpeterson: Yes, probably something else that warrants a comment, and/or updating the top level doc string. Starting with v1.0.8 the worker should always upload the metrics outputs

roberto-araiza-curibio (approved):

jperson (approved):

PR #867, opened 2024-09-03
Initial notifications endpoint

roberto-araiza-curibio:


deployments/pulse3d/services/pulse3d/src/models/models.py
```diff
@@ -7,6 +8,18 @@
...
+class NotificationType(StrEnum):
+    CUSTOMERS_AND_USERS = auto()
+    CUSTOMERS = auto()
+    USERS = auto()
+
+
+class SaveNotificationRequest(BaseModel):
+    subject: str
+    body: str
+    notification_type: NotificationType = NotificationType.CUSTOMERS_AND_USERS
```
tannermpeterson, line 20: You might need to use `notification_type: NotificationType = Field(default=NotificationType.CUSTOMERS_AND_USERS)`. I can't remember off the top of my head what the difference is though

tannermpeterson (approved):

deployments/pulse3d/services/pulse3d/src/models/models.py
roberto-araiza-curibio: I just double-checked and I think this is OK.

According to the [docs](https://docs.pydantic.dev/latest/concepts/models/#basic-model-usage), this is the straightforward way to set a string default.

We can use the `Field(...)` function to also define [defaults](https://docs.pydantic.dev/latest/concepts/fields/#default-values), and other customizations and [constraints](https://docs.pydantic.dev/latest/concepts/fields/#string-constraints) (e.g. max_length for strings).

For simple string defaults it looks like both approaches are equivalent, so I'm inclined to leave this as-is.

PR #883, opened 2024-10-01
Add DB trigger to notify of INSERTs on notification_messages table

roberto-araiza-curibio:
---
  - https://app.asana.com/0/0/1208442213650612

### DIFF
--- a/core/db/curibio/alembic/versions/4319d72ea635_notificationmessagesnotifytrigger.py
+++ b/core/db/curibio/alembic/versions/4319d72ea635_notificationmessagesnotifytrigger.py
@@ -1,0 +1,53 @@
+"""NotificationMessagesNotifyTrigger
+
+Revision ID: 4319d72ea635
+Revises: a869e370f0c7
+Create Date: 2024-10-01 12:11:07.904287
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "4319d72ea635"
+down_revision = "a869e370f0c7"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        """
+        CREATE OR REPLACE FUNCTION notification_messages_notify()
+        RETURNS TRIGGER AS $$
+        BEGIN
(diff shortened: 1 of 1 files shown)


tannermpeterson (approved):

core/db/curibio/alembic/versions/4319d72ea635_notificationmessagesnotifytrigger.py
```diff
@@ -0,0 +1,53 @@
...
+        CREATE OR REPLACE FUNCTION notification_messages_notify()
+        RETURNS TRIGGER AS $$
+        BEGIN
+            PERFORM pg_notify(
+                'events',
+                (
+                    json_build_object(
+                        'table', 'notification_messages',
+                        'subject', (SELECT subject FROM notifications WHERE notifications.id=NEW.notification_id),
+                        'body', (SELECT body FROM notifications WHERE notifications.id=NEW.notification_id)
```
jperson, line 30: will this contain the full message + subject in the notify event?

 If so it might cause problems because the payload has a max length of 8000 bytes in the default configuration, see the parameters section in https://www.postgresql.org/docs/13/sql-notify.html.  Not sure if we can change that configuration since we are using aurora rds.

tannermpeterson: Yeah, the way we've gotten around this limit with the other notifies is have the payload contain the minimum amount of info necessary and have the event broker grab anything else it needs from the DB after it receives the payload. In this case the event broker could grab the subject and body col from the notifications table as long as the notification id is in the payload

roberto-araiza-curibio: Thanks for pointing this out, guys.  I guess that explains the "SELECT meta..." queries I saw in the event-broker.  I was trying to minimize DB trips but it looks like that's the way to go to avoid the payload max length constraint.

I'll close this PR and send another one with the better approach.

PR #891, opened 2024-10-09
Create endpoint to mark notification_messages as viewed

roberto-araiza-curibio:
Example running the same request twice:

First run
Screenshot 2024-10-09 at 12 04 42 PM

Repeat
Screenshot 2024-10-09 at 12 06 08 PM

---
  - https://app.asana.com/0/0/1208494116369388

### DIFF
--- a/deployments/pulse3d/services/pulse3d/src/main.py
+++ b/deployments/pulse3d/services/pulse3d/src/main.py
@@ -968,7 +968,7 @@ async def get_notification_messages(
 
 @app.post("/notification_messages", response_model=ViewNotificationMessageResponse)
 async def view_notification_message(
-    view_request: ViewNotificationMessageRequest, token=Depends(ProtectedAny(tag=ScopeTags.PULSE3D_WRITE))
+    view_request: ViewNotificationMessageRequest, token=Depends(ProtectedAny(scopes=list(Scopes)))
 ):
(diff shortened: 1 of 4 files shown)


deployments/pulse3d/services/pulse3d/src/main.py
```diff
@@ -964,6 +966,22 @@ async def get_notification_messages(
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
 
+@app.post("/notification_messages", response_model=ViewNotificationMessageResponse)
```
jperson, line 969: since this is updating a row should it be a put instead of a post?

jperson: actually, since this method isn't idempotent maybe it makes more sense as a post

roberto-araiza-curibio: This one is a tough one to decide, and I always get them wrong anyways.   The [mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) say:

> The POST method submits an entity to the specified resource, often causing a change in state or side effects on the server.
> The PUT method replaces all current representations of the target resource with the request content.


I'm inclined to leave this as a POST.

jperson (approved):

PR #916, opened 2024-11-25
Fix preset deletion cancel

tannermpeterson:
Also:
- Tweak widths of notifications creation table to fill full page
- Add read/unread col to notifications message table. Table did not seem to want to let the "subject" col grow larger than 1000px, so I added this col to fill the remaining width. Users can sort by this col to quickly bring all unread notifications to the top


roberto-araiza-curibio (approved):
Thanks for fixing the notification columns!

